PR #141, opened 2025-04-16
Lore pirate edition

LukeZurg22:
Not sure if I missed a spot, but we'll find out.

:cl:
- tweak: Adjusted Pirates -> Rogue naming.


LukeZurg22:
There are some deliberate neglected changes from pirate -> rogue. Namely for pirate victorian-age weapons, which are very pirate-y.

PR #158, opened 2025-04-18
SNS Amicitia

LukeZurg22:
:cl:
- add: Amicitia ship (by Game_and_Glitch)


starch70 (changes requested):
Why does this remove a random changelog. Also why is it using friendships for EMP launchers. Make a dedicated ship gun for EMP launchers. Shields will not block projectiles fired by friendships. And please add some kind of decals and decoration.. This looks rushed and flat.

LukeZurg22:
> Why does this remove a random changelog. Also why is it using friendships for EMP launchers.

Great question, not sure. As for the rest:
 Make a dedicated ship gun for EMP launchers. Shields will not block projectiles fired by friendships. And please add some kind of decals and decoration.. This looks rushed and flat.
- I'll pass it on.

LukeZurg22:
The ship tbh wasn't very well put together from the start. Plus wacky commit shenanigan's are afoot here that i'd rather just have it be fresh from a fresh branch.

PR #249, opened 2025-04-26
Fentanyl Balance. Actually balanced (TM)

LukeZurg22:
I've played Chem and Medic long enough to know the recipe is still **very borked.**

:cl:
- fix: The Fent Crisis, hopefully.


LukeZurg22:
Making Fent Contraband as well.

PR #294, opened 2025-04-30
Removed the ability to find over 100 TC after a handful of wrecks.

LukeZurg22:
PR's should be balance tested before merge whereatall possible.

:cl:
- fix: No more 100 TC up per 6-10 wrecks.


SupernoobTheN1:
Huh? Wrecks? These should only spawn on dead drops, that no one does

PR #379, opened 2025-05-07
Fix Compatibility Mode

gusxyz:
## Why / Balance
People with older computers may not be able to play without this enabled, good idea to support it.
easy fix issue was the shader was setting the value while defining the uniform which compatibility mode does not support has to be done inside the fragment method.

## Requirements

- [ ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ ] I have added media to this PR or it does not require an ingame showcase.

## Breaking changes

**Changelog**

:cl: gus
- fix: Compatibility mode will work once again.


ark1368 (approved):
Tested and it works. Thank you for contributing! 👍

PR #469, opened 2025-05-14
Flashpoint update

core-mene:
fixed missing shield computer
updated description


Junerisms:
Apparently the shield generator doesn't work with it's original spawn either. Might've been a one time issue, swap the spots? :clueless: **(when this was reported in an ahelp my fix was just deleting the shield generator they had and replacing it with a new shield generator entity. then after linking it worked just fine. Same exact spot aswell.)**

core-mene:
i dunno when i tested on local it worked fine after linking

core-mene:
image
:clueless:

SolNostra:
Tested, works fine.

PR #471, opened 2025-05-14
Flashpoint update V2

core-mene:
after playtesting it found out the dispo pipe led to inside a wall + it was missing defibs
this SHOULD or maybe will i dunno fix it idfk


SolNostra:
You post one more flashpoint update I'm KILLING you.

PR #488, opened 2025-05-16
Increases radio ranges to 6k.

Redrover1760:
## Why / Balance

Radio is too disconnected/lonely, to the point where calling for help over comms or trying to chat with people, etc. Is mostly just not done at all. Sure, there are other methods to do these things, but even trying to actively patrol out and about as TSF won't help much as the radio ranges are so short most people won't even try.

Making it 6k or 5k radius would still have scenarios where there is simply no help in the area, especially if you venture out far, but still would allow more safety and more ability for TSF ships to patrol around and talk to people. This also means that you can talk to someone who just FTL'd away over radio, or hear an issue or request and FTL to it, radio being about the distance of FTL makes sense. Should help with people asking to join a crew as well.

Makes playing TSF feel very disconnected with everyone else. Same with ships of the same companies trying to cooperate with each other.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Increases radio range to 6k.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Devices/encryption_keys.yml
+++ b/Resources/Prototypes/Entities/Objects/Devices/encryption_keys.yml
@@ -18,7 +18,7 @@
   parent: [EncryptionKey, RecyclableItemDeviceTiny] # Frontier: added RecyclableItemDeviceTiny
   id: EncryptionKeyCommon
   name: broadband encryption key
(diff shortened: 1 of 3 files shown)


starch70 (changes requested):
update guidebook page with new range too

Redrover1760:
> update guidebook page with new range too

Done

ark1368 (approved):

PR #500, opened 2025-05-18
New Ruleset

ark1368:
## About the PR

Thought the current ruleset could use a rework with the same ideals in mind.

## Why / Balance

Would hopefully reduce NRP.

## How to test

Reading.

## Requirements

- [X] I have agreed to turn this into a hugbox utopia.

**Changelog**

:cl: Ark
- tweak: A new ruleset has arrived. It aims to curve the NRP behavior while staying minimal and prioritizing player freedom.

### DIFF
--- a/Content.Shared/CCVar/CCVars.Server.cs
+++ b/Content.Shared/CCVar/CCVars.Server.cs
@@ -18,7 +18,7 @@ public sealed partial class CCVars
     ///     Guide Entry Prototype ID to be displayed as the server rules.
     /// </summary>
     public static readonly CVarDef<string> RulesFile =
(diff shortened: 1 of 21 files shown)


Redrover1760:
suggestions: rename combat rules to conflict rules

mention that new life rule applies to respawns and not revivals. 

otherwise, I like it already.

Redrover1760:
Roleplay Rule 5 is an interesting concept, although what "emotes" involves/means could be confusing (especially to newer players or those that aren't familiar with HRP level servers and stuff), and it feels like it would be better if it was written in a more positive "Seek to X" manner instead of restrictive "Do not"

I'll reference one of Monkestation's RP Rule/Guide things, a fairly popular SS13 server here who has a fairly similar guideline/rule of sorts https://wiki.monkestation.com/en/info/rules#yes-and

Yes And

PR #581, opened 2025-05-25
2 new ship guns (spritted and collision box) + cyrexa resprite

core-mene:
Respritted cyrexa (bigger cannon)
Added the Kargil 203mm cannon
Added the Bofors 255mm cannon
image


BlueHNT:
While I like your changes with the fixtures. I highly recommend setting the fixtures up around the base while ignoring the cannon itself. Mainly because it will turn around and could create some issues due to that. So setting up the old one around the base would most definitely be a better choice.
image

BlueHNT:
Other than my nitpicks, this is a great idea for a variety of similar weaponry!

core-mene:
updated their sprites + updated hitbox + updated states

ark1368 (approved):

PR #775, opened 2025-06-10
Fixes the orientation indicator when using FTL

RikuTheKiller:
## About the PR

Makes it so when you use the scroll wheel, the ship pre-visual actually represents your final orientation.

Previously, it was mirrored over the X-axis, this had to do with the coordinate space of textures (X right, Y down) being different from the coordinate space of the world (X right, Y up), leading to RotateVec doing a clockwise rotation instead of counter-clockwise like it should.

This PR has been tested in-game.

## Why / Balance
I want to see where I'm going damnit.

## How to test
Just use FTL and see if the orientation is right.

## Requirements

**Changelog**

:cl:
- fix: The FTL orientation indicator actually points in the right direction, rather than being mirrored over the X axis.


mikusssssss (approved):
the wonders of writing a single symbol wrong

ark1368 (approved):

PR #820, opened 2025-06-14
[EXPERIMENTAL] Makes all shipgun projectiles around 3 times faster and fixes missile tracking

RikuTheKiller:
## About the PR
Title, all projectiles fired from shipguns go about 3 times faster now.
Also the scan arc property of missile tracking was completely fucked, so I fixed that.
This means missiles will no longer swerve 90 degrees to hit your allies in the ass.
Oh also ranges have been tweaked so charon projectiles don't go 13500m anymore.

Tested, the changes do work but I don't know how they'll play out in live fights.

## Why / Balance
I'm tired of my gun paths looking like wet noodles within 300m of my target.
This is also intended to make missiles less of an universal pain in the ass with FTL.
Gadyukas for example are incredibly mobile and can just dodge any and all gunfire.
It's not particularly fair if the only counter to missiles is to also use missiles. (so missile ships counter 90% of other ships)

## How to test
Fire every gun in the game, if you want to be thorough.
But ideally test this on live, which is why it's "experimental".

## Requirements

**Changelog**

:cl:
- tweak: [EXPERIMENTAL] All shipgun projectiles are 3 times faster, ranges have been tweaked and missile tracking works better. Report any and all issues you have with this change, it's up for debate and needs live testing.


ark1368 (approved):
Experimental :trollface:

Redrover1760:
yeah this, this really screws over smaller ships. Missiles were unbearable to deal as a smaller ship but now dodging anything is nigh impossible.

Mobility should be a damage reduction factor and 3x speed makes it not very significant if you can aim at all. I do like the speed increase but 3x more seems too much from my first impressions.

also highlights Tarynxs as absurdly unbearable.

PR #889, opened 2025-06-20
Marauder buff

Nope1356:
Plasma cannon that is wildly underperforming in practice compared to every other option in its weight class. Requires multiple to be effective in its current state, however this buffs them to be more generally effective.

This makes the marauder hit a bit harder than it typically would. Depending on how it performs in the game (not in a test enviroment) I will decrease the fire rate to balance it.

:cl:
- tweak: TheyHeldMeDownAndForcedMeToType Buff of Marauder to be less shit more good.

### DIFF
--- a/Resources/Prototypes/_Mono/Entities/SpaceArtillery/SpaceArtillery/Energy/projectiles.yml
+++ b/Resources/Prototypes/_Mono/Entities/SpaceArtillery/SpaceArtillery/Energy/projectiles.yml
@@ -10,7 +10,7 @@
     damage:
       types:
         Radiation: 1250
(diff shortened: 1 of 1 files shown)


mikusssssss (changes requested):
changelog changelog changelog changelog changelog changelog changelog

BlueHNT:
I've gone over an extensive list of the 3 medium guns in mono and made a fuller buff to match it with other guns. https://github.com/Monolith-Station/Monolith/pull/891

Not sure if this mini-buff is viable after the fixes.

ark1368:
#891

PR #897, opened 2025-06-21
The Performance PR

sleepyyapril:
## About the PR

This PR does some miscellaneous warnings fixes and also makes sure that mobs don't try to pathfind when there is no one remotely nearby. (>= 2000)

I should probably move the nearestplayer thing to pathfindingsystem or something, but whatever.

## Why / Balance

You have a lotta mobs. This is a stopgap.

## How to test

Run PR in debug
Teleport yourself far away
Have another client watch in ghost
They don't move
Get within 2000 units
They move

## Media

https://github.com/user-attachments/assets/7320f7a6-dd5a-4dc4-b725-40f0182181f3

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.
- [x] I can confirm this PR contains no AI-generated content, and did not use any AI-generated content.

**Changelog**

:cl:
- fix: Maybe possibly performance fixed a little. I dunno, I just work here.

PR #905, opened 2025-06-22
A Performance PR

sleepyyapril:
## About the PR

reverted depth changes
trying #897 again, but:
- don't set when closest is null, we only care about distance, and that still exists
- remove more checks we don't need

### DIFF
--- a/Content.Server/Worldgen/Components/Debris/DebrisFeaturePlacerControllerComponent.cs
+++ b/Content.Server/Worldgen/Components/Debris/DebrisFeaturePlacerControllerComponent.cs
@@ -27,7 +27,7 @@ public sealed partial class DebrisFeaturePlacerControllerComponent : Component
     /// <summary>
     ///     The chance spawning a piece of debris will just be cancelled randomly.
     /// </summary>
-    [DataField("randomCancelChance")] public float RandomCancellationChance = 0.25f; // Mono 0.35f->0.85f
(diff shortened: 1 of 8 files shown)


ark1368:
the-here-we-go-again-meme-is-perfect-for-every-frustrating-s_kgv8 1200-1475340496

sleepyyapril:
dont merge, making the "grid nearby check" a cvar

Content.Server/Worldgen/Components/Debris/DebrisFeaturePlacerControllerComponent.cs
```diff
@@ -27,7 +27,7 @@ public sealed partial class DebrisFeaturePlacerControllerComponent : Component
     /// <summary>
     ///     The chance spawning a piece of debris will just be cancelled randomly.
     /// </summary>
-    [DataField("randomCancelChance")] public float RandomCancellationChance = 0.85f; // Mono 0.35f->0.85f
+    [DataField("randomCancelChance")] public float RandomCancellationChance = 0.35f; // Mono 0.35f->0.25f, but larger radius
```
Redrover1760: comment doesn't line up with code

sleepyyapril: i forgor

sleepyyapril:
okay. we good.

ark1368 (approved):
:trollface:

PR #909, opened 2025-06-22
Asteroid dead zone and sparse zone around Frontier (#3112)

sleepyyapril:
https://github.com/new-frontiers-14/frontier-station-14/pull/2910

:cl: sleepyapril, whatston3
- add: Ported Frontier's asteroid belt debris changes


Redrover1760:
just to note, we doubled all of frontier's distances at some point. POI distance, etc, so dead zones/biomes would have to be doubled or something like that too

mikusssssss (approved):
this is peak but wheres the changelog

ark1368 (approved):

PR #1006, opened 2025-06-29
Ship Access Update

ark1368:
## Requirements

- [ ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ ] I have added media to this PR or it does not require an ingame showcase.
- [ ] I can confirm this PR contains no AI-generated content, and did not use any AI-generated content.

## Breaking changes

**Changelog**

:cl: Ark
- add: Ships are now restricted to ID cards with deed access. You can still grant guest access by allowing others to swipe your unlocked shuttle console, without providing them a copy of your deed.

PR #1023, opened 2025-06-29
Revert "Ship Access Update"

Redrover1760:
Reverts Monolith-Station/Monolith#1006

PR causes quite a few major issues and this needs to be looked over again.

Of note:

Vouchers do not work anymore, you cannot access doors of a ship you bought a voucher of. No one can.

Cyborgs cannot gain guest access to a ship at all, so they just can't play the game.

Faction ships have ship guest access mechanics too, meaning despite having TSFMC access you can't board and save your ally. Not to mention it makes boarding very difficult in general. May need to give every major faction access breakers at this rate, and breaking into your ally's ship to save them isn't really ideal at all.

Paramedics without breaching tools is agony.


UnicornOnLSD:
system is good just lacked taking those things into account.

Boarding isn't that much of an issue in my eyes, basic ship security's great.

For factions though, yeah it's a major issue.

paramedics alike.

PR #1126, opened 2025-07-04
Anchors Halcyon

Redrover1760:
## About the PR

yeah so anyways that was a fun round lets not do that again

I will look into later making only gyros work on faction POIs so they can rotate guns and shit for better firing accuracy and resisting boarding but for now we're going back to standard monolith gameplay (tm)

## Why / Balance

Maybe there's a world where this works, but I don't think we've reached a balance state where we can have this sorta thing be what it is. 

Flyssa should be more akin to the walking juggernaut with IFF locked on, and that is actually fairly doable in that respect.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Halcyon not potentially mobile anymore.


BoskiYourk:
The round was fun, but it was a bit much. Good change!

PR #1281, opened 2025-07-09
CloakHeatSystem only Applies Heat when Moving

ScyronX:
## About the PR
CloakHeatSystem now only applies heat when vessel is moving

**Changelog**

:cl: Scyron
- tweak: Changed Prowler cloak only heats when moving!
- tweak: Prowler cloak heat capacity has been reduced to 60 seconds!


starch70 (changes requested):
needs to get MaxCloakTime too in CloakHeatComponent

starch70 (approved):

PR #1320, opened 2025-07-11
Removes Magnets From Generators

Redrover1760:
## About the PR

magnets are laggy, but these are especially laggy as they'd just keep trying to insert every single piece of fuel nearby nonstop every second despite not having room. At least other magnets insert the materials and stop.

## Why / Balance

lag. They have 100 storage now so there's no reason for it to have a magnet.

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Magnets removed from generators. Lag.


Danchi299:
Step 1. Increase generator capacity you can refuel 100 at a time with magnets
Step 2. Remove magnets
Bravo, Brilliant, Outstanding.
At this point we should make fuel cap something ridiculous like 1_000_000 and never use magnets again.

PR #1321, opened 2025-07-11
Actually Optimizes Magnets

Redrover1760:
## About the PR

why have 500 local timers for each individual magnet that forces you to call EntityQueryEnumerator every single tick when you can have a single global timer and check out every magnet and call EntityQueryEnumerator at once per second?

instead of calling every single magnet every single tick to see if its on and that its activated and if its time yet for it to magnetize things

also comp.NextScan < currentTime instead of currentTime < comp.NextScan for some ungodly reason when NextScan is the thing being incremented

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Magnet optimizations


starch70 (approved):
Goidamerge

PR #1322, opened 2025-07-11
Revert "Actually Optimizes Magnets"

Redrover1760:
Reverts Monolith-Station/Monolith#1321

Ok I'm not smart


Danchi299:
So... what happened?

PR #1487, opened 2025-07-19
Partially Unmald Plant Bag

tonotom1:
## About the PR

Partially walks back a plant bag nerf made on Frontier. It doesn't return the same level of power because a 1x2 storage of medicinal plants is easily stronger than a bucket. Instead it is 2x2.

## Why / Balance

Walking back the nerf re-enables the ability of botanists to carry their catalogue of seeds and samples to share without sacrificing a belt slot. It's not the same use case as a construction or ore bag. They're used to bring things from a garden, not extract materials from a roid/wreck.

If you think you're clever bringing medical plants in a bag, the 2x2 size makes the bucket competitive enough to not make plant bags a braindead decision. You can grind plants (50 potency factored in) and compress more than a plant bag's worth into a bucket and also inject the chems way faster in a hypo. With potency beyond 50 and multiple chem hybridization one could surpass a bucket but perhaps botanists that can reliably breed high volume fruits with actually usable chemical compositions deserve _some_ take-along power?

Bluespace plant bags have been considered and nothing has been done to address them: bluespace beakers also exist and once again mog the front of storage at its 2x2 size. It's 8x liquid storage (in two beakers) vs 5x fruit carry capacity. With that out of the way, let botanists have fun.

## Media

image

## Requirements

- [X ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X ] I have added media to this PR or it does not require an ingame showcase.
- [X ] I can confirm this PR contains no AI-generated content, and did not use any AI-generated content.

## Breaking changes

i touched yaml

**Changelog**

:cl:
- tweak: Plant bags can once again be stored in bags. Gardeners can peddle their produce on foot more.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Specific/Hydroponics/tools.yml
+++ b/Resources/Prototypes/Entities/Objects/Specific/Hydroponics/tools.yml
@@ -144,7 +144,7 @@
     # End Frontier
   - type: Item
     storedRotation: -90
-    size: Medium # Frontier
+    size: Normal # Mono
   - type: Clothing
     quickEquip: false
     slots:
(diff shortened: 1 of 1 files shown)


Redrover1760:
Another frontier holdover gone

Redrover1760 (approved):

PR #1858, opened 2025-08-11
Retires the Parabellum

Redrover1760:
## About the PR

Removes the Parabellum.

## Why / Balance

The ship doesn't really fit the security officer gameplay loop and has no utility value. Sec ships should have high utility value first, a larger sec ship should be some kind of, well armed permabrig or security dropship carrier or the like instead of a flat emp warship.

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Retires the Parabellum


Redrover1760:
> I think we should keep w/ edits it until we have more sec ships atleast, the only 2 other options for secguard ship if we get rid of it is a detective ship and the horribly outdated arachnid. CO service shipyard is already very devoid of ships, and most of the ones there are very low-quality.

got it

PR #1933, opened 2025-08-17
Medical Dispatch Rework

Vinny-Shick:
## About the PR
Remade MD to be more functional and to be  more of a social hub instead of  a place you buy a ship from and return body bounties to 

## Why / Balance
Current MD is severely lacking in its design and function for MD staff 
## How to test

Run the file 
## Media
image
image

## Requirements

## Breaking changes

**Changelog**

<--
:cl:
- Tweak: Remade Medical dispatch 
-->


Sarahon:
Ngl, I don't really understand the logistics of this medical dispatch. Let me break it down.

Imagine you were to steer to the opposite side of the medical room, they now have a 2 man camel ride to drag their gasping dying friend to the medical area. 

You'll need to label it well. But I don't quite understand the shape. You'll need to explain it to me.

The docks being so close means big fat ships will be able to block all 3 of them

Vinny-Shick:
> Ngl, I don't really understand the logistics of this medical dispatch. Let me break it down.
> 
> Imagine you were to steer to the opposite side of the medical room, they now have a 2 man camel ride to drag their gasping dying friend to the medical area.
> 
> You'll need to label it well. But I don't quite understand the shape. You'll need to explain it to me.
> 
> The docks being so close means big fat ships will be able to block all 3 of them

image

this is the shape

also the logistics are quite simple  it has 4 wings for 4 different purposes with all of them being accessible form the centre

grandalff:
Well done! Extremely well done.

ThanosDeGraf:
I do like the central placement of the medical area. It was weird how that was placed like an afterthought, but with this it's fully kitted out (^^ ). Besides that, the cardinal star shape is a fantastic idea, if only the station would always spawn pointing north correctly.

My only concern is as Sarahon said, that big and small ships could easily block each other from docking up, which can be problematic whenever its payout time and all of dispatch arrives.

One thing I would like to ask is if you could make a screenshot of the station while it has a bunch of small, medium and two large type MED ships docked?

Junerisms:
I think medical dispatch does need a rework, and I like the idea of this one!

### However, I am not a fan of the snowflake design. ### 

1. There aren't as many docks as you'd like to presume there to be, only taking up a rough estimate of 12 shuttles (and thats if people dock respectfully which how long and protruded these docks are)
2. The hallways from the docks are **brutally** expansive and claustrophobic. Starting from a single tile before expanding outwards. Makes the pain of building shuttles very prominent, and it only respects **one way traffic.** If they're carrying a box you're going to be forced out of the way and you must wait your turn.
3. Regarding the same hallways issue, this would be a drag for Emergency Responders to haul across this entire gigantic Dispatch. (Look at the medical wing for example, the cryogenics is so far deep into the left side of the Snowflake/Compass)
4. Security having a bigger spot at Medical Dispatch is cool, although definitely overzealous as you focused more on the murdering part of security than the actual security part. (there is only one brig.)
5. Director of Care office is in a decrepit corner right next to Security with no windows. (except for the odd bend at the back) Looks more backalley and drug dealing than an official working under the Trans-Solar Federation.

OneZerooo0 (commented):
Function > form
I agree with what June says about this. It’s clearly designed with the shape more in mind than logistics and could use some change.
Also, the shape just isn’t very interesting - Stars are cool but vary the hull form more

Skyguard12:
I have a question: Why is there an armory in a hospital?

Redrover1760:
This one isn't it, sorry. If you can somehow improve its current state you are welcome to make a new pr.

PR #1939, opened 2025-08-18
Buff Ptk

Ilya246:
## About the PR
ptk:
15 -> 40 blunt
15 -> 400 structural

## Why / Balance
makes it do actual damage because it doesn't right now
it has like no range and no projectile speed and low-ish firerate so this is balanced
if this somehow ends up being an overbuff just nerf it later

## How to test
fire ptk at wall
watch it actually do damage

**Changelog**

:cl:
- tweak: PTK-1500e: damage 15->40 blunt, 15->400 structural


OneZerooo0 (approved):

DoomAngelBlade:
Now there will be ships, sole purpose of which is to find a blind spot, stick to ship and drill in.

burrrkeye:
Ptk flatpack in my boarding loadout instead of C4

Ilya246:
> Now there will be ships, sole purpose of which is to find a blind spot, stick to ship and drill in.

yes good

PR #1957, opened 2025-08-19
HES-Canister, an Entry Level Gas Mining Ship 

Rubeebeebee:
## About the PR

The Canister is an entry level gas mining ship, capable of harvesting filtered gas from deposits, with it's default setting made to harvest nothing but plasma, to ensure the highest profit ratio out of raw gasses. 

It's got the barebones for a gas mining ship, lacking even shipbound mining guns, relying on PKAs, pickaxes, and whatever the end user has manufactured from the salvage techfab. 

## Why / Balance

It's a cheap entry level gas mining ship, it aims to be the replacement for the akasono (since franky, the ergonomics of the akasono as a gas mining ship was not good).  Hopefully this ship would be a better introduction to the industry of gas mining for new and curious players. 

## How to test

load up the .yml in an IDE or buy it from CIV shipyards

## Media

Canister Profile Live
Canister Profile
canister appraisal
canister warp and latejoin

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added the HES Canister, an entry level gas mining ship


OneZerooo0:
We already have enough gas mining ships, sorry

Rubeebeebee:
> We already have enough gas mining ships, sorry

aw man! can i argue that this could be a decent replacement for the akasono? it's a cheapo entry level gas mining ship for curious players

starch70 (changes requested):
Trim decals on the interior are a bit bland and samey, try overlays/tech too
Gas chamber should use reinforced (either normal or hardened) tiles, hatch should be swapped for external glass airlock
Could have more space inside if grilles covering thrusters were taken off, not really a combat ship that needs it
Docking port diagonals need to be full walls
Extending the gaslock (extraction one) w/ a probe might be useful
Needs a bit of LV cables & steel at the front for wiring/piping/lattice for gas mining setup
Gyro and gravgen should be accessible w/o needing decon
No gas deposit scanner anywhere as far as i can see

Rubeebeebee:
> Trim decals on the interior are a bit bland and samey, try overlays/tech too Gas chamber should use reinforced (either normal or hardened) tiles, hatch should be swapped for external glass airlock Could have more space inside if grilles covering thrusters were taken off, not really a combat ship that needs it Docking port diagonals need to be full walls Extending the gaslock (extraction one) w/ a probe might be useful Needs a bit of LV cables & steel at the front for wiring/piping/lattice for gas mining setup Gyro and gravgen should be accessible w/o needing decon No gas deposit scanner anywhere as far as i can see

canister updated live
canister updated
canister updated atmos network

Right! i've updated it with your requests; 

The floors and decals have been overhauled, ive went with catwalks on the exterior so the end user can see the wires underneath, and it adds to the industrialpunk:tm: feeling of the ship, adds to the extremely utilitarian vibe, i think. The entrance has this little "pit" where it gives the illusion of depth of going down some steps and going back up

Gas chamber floor has been replaced with reinforced tiles, and ive added decals on the inside to help identify which vent does which. 

Docking ports are no longer diagonals!

I've extended the gaslock a bit! 

canister gas deposit scanners + LV wires
For the materials and gas deposit scanner, they were in the materials wall locker (gas deposit scanner, electronics toolbox, and the steel that the locker comes with), but i've added a second gas deposit scanner out in the open, just in case. 

Gyro and gravgen is now accessible without decon, just replaced the windows with windoors. 

@starch70 Thank you for reviewing! You've been very poggers

Rubeebeebee:
also the table with the microwave and bedroll seems like what workers would bring onto the ship, since it mightve not come with it originally, which gives it a nice touch of "lived in"

Rubeebeebee:
@starch70 THANK YOU!!! WOOO!!

PR #1968, opened 2025-08-20
Any Ship Expedition Console

Ilya246:
## About the PR
- expedition console can now be used on non-exped ships
- added exped console flatpack to expeditionary flatpackvend

## Why / Balance
why did they only work on exped ships in the first place
just let any ship exped

## How to test
buy console flatpack at expeditionary flatpackvend, put on non-exped ship, do exped

## Media
image

image

**Changelog**

:cl:
- add: You can now buy expeditionary console flatpacks at the expeditionary flatpackvend.
- tweak: Expeditionary console now works on non-expedition ships.


Borgousse:
MONOLITH, MERGE THIS PR AND MY LIFE IS YOURS

Danchi299:
Does this mean all EXP model ships move to CIV and MIL classes?

Ilya246:
> Does this mean all EXP model ships move to CIV and MIL classes?

actually makes sense but should be PR'd separately i guess

LittleNorthStar:
I mean even if the expedition feature is no longer unique in terms of combat capability (in both ship and crew) they still sit between CIV and MIL so the label is not entirely superfluous, however it could be changed to something like PMC to better represent they no longer have a monopoly on the niche of expeditions but are still more formidable than regular civilians

fow111:
another w from peaktributor

LaCumbiaDelCoronavirus:
good but i think it should be (easily) researchable rather than buyable, to keep incentive of either buying a sci/exped ship or stealing it

Ilya246:
> good but i think it should be (easily) researchable rather than buyable, to keep incentive of either buying a sci/exped ship or stealing it

things like this should generally be buyable
also people will just buy duran steal the console and sell it if i do this

EckoAurum:
I like the change, makes things interesting.
However, as much as I would love for this to be the case, you may want to add exceptions.
Lest the people take their favorite stations for a joyride 
image

AveriV2:
> I like the change, makes things interesting. However, as much as I would love for this to be the case, you may want to add exceptions. Lest the people take their favorite stations for a joyride

keep it. atleast for a day or two.

Ilya246:
> I like the change, makes things interesting.
However, as much as I would love for this to be the case, you may want to add exceptions.
Lest the people take their favorite stations for a joyride

1984'd

ScyronX:
Put a size limit on it at least so at least the ship is bigger then 3x3

Ilya246:
> Put a size limit on it at least so at least the ship is bigger then 3x3

why

ScyronX:
> > Put a size limit on it at least so at least the ship is bigger then 3x3
> 
> why

because small ship functionality has been buffed enough from the new mass scanner system to the thruster size max speed limit changes I think we are reaching a point at which baegs are better and more preferable to literally all other ships due to their stealth mobility and functionality

Danchi299:
> > > Put a size limit on it at least so at least the ship is bigger then 3x3
> 
> -a point at which baegs are better and more preferable --- due to their functionality-

Baeg is a 4x4.
Making a 3x3 limit would do nothing of value besides denying the silly fun of 2x2 and 1x1 ships, which are more of a handicap than actual meta strategy, as only upside is ship traveling faster in FTL.
So either make it around Duran size limit (~50 tiles) or don't make it at all.
Unless we are trying to deny a 2x2 Charon + Exped + Warp Drive + Shuttle Console play for quick FTL out?

Ilya246:
> because small ship functionality has been buffed enough from the new mass scanner system to the thruster size max speed limit changes I think we are reaching a point at which baegs are better and more preferable to literally all other ships due to their stealth mobility and functionality

i think that's okay and no cargo space
if we want to change this for some reason, should be addressed separately and not just random "no can't do that on small ship"

Arcticular1:
Maybe a good compromise in terms of balance would be making it so FTL and exped consoles can’t be used on the same ship? Limits it in a different way opposed to giving every ship a feature with no drawback

Ilya246:
> Maybe a good compromise in terms of balance would be making it so FTL and exped consoles can’t be used on the same ship? Limits it in a different way opposed to giving every ship a feature with no drawback

i think it's fine and we're just making an issue from nothing

Arcticular1:
> > Maybe a good compromise in terms of balance would be making it so FTL and exped consoles can’t be used on the same ship? Limits it in a different way opposed to giving every ship a feature with no drawback
> 
> i think it's fine and we're just making an issue from nothing

I can see abuse cases for this. Specifically in faction gameplay.

Ilya246:
> I can see abuse cases for this. Specifically in faction gameplay.

it has same restrictions as normal FTL, i still think we're inventing theoretical problems that won't actually happen
let's just try it and see

PR #1982, opened 2025-08-21
Fix Deedless Shuttle Dampening

Ilya246:
## About the PR
- fixes dampening controls not showing up for shuttles without a deed
- removes DeedlessShuttleComponent because with this it will just do nothing

this makes it so that loadgrid'd shuttles can have dampening set
and shuttles built from zero
and no there's no way this can cause you to be able to unanchor a grid because grid anchoring relies on physics bodytype not dampening

## Why / Balance
fix

## How to test
loadgrid shuttle
observe it actually have dampening controls

## Media
image
loadgrid'd baeg

**Changelog**

:cl:
- fix: Fixed grids without a deed not having dampening controls on their shuttle console.


Redrover1760 (approved):
ty past self was not very smart with fixing this issue

PR #1983, opened 2025-08-21
Movable Vgroid

Ilya246:
## About the PR
- you can now move and pilot the vgroid
- artifacts can now activate on vgroid

makes more sense with #1981 and works better with #1982

## Why / Balance
fun
it's probably fine since it requires an absurd amount of thrusters to move anyway
if people somehow abuse this just give them a decreased max speed or something
i tested ramming it into a poi with a lot of thrusters and it doesn't do much damage

## How to test
put shuttle console and like 30 thrusters and 30 gyros on vgroid
fly

## Media
\<image of vgroid flying\>
tested, works

**Changelog**

:cl:
- tweak: VGroids are no longer static and may be moved.
- tweak: Artifacts can now activate on VGroids.

### DIFF
--- a/Resources/Prototypes/_NF/Events/nf_bluespace_dungeons_events.yml
+++ b/Resources/Prototypes/_NF/Events/nf_bluespace_dungeons_events.yml
@@ -44,6 +44,7 @@
           inherent: true
         - type: IFF
         - type: LinkedLifecycleGridParent
+        # Mono - make it movable
         protos:
         - NFVGRoidBasalt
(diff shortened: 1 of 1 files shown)


OneZerooo0 (changes requested):
Comment out the removed features instead of deleting them

Ilya246:
> Comment out the removed features instead of deleting them

done

Redrover1760 (approved):
godo

PR #1987, opened 2025-08-21
Banner Grid Claiming

Ilya246:
## About the PR
- you can now altclick banners to "claim" debris/event grids
- claimed grids will not despawn
- this way you can make permanent base on a vgroid or UIV or debris or etc

## Why / Balance
good
if people use this too much and cause lag just limit it in some way, but i don't think it'll happen so it's probably fine for now

## How to test
put banner on rock, anchor, altclick, fly 10km away from rock and back, watch rock still be there
altclick banner again, fly 10km away from rock and back, rock gone

put banner on vgroid, anchor, altclick, end vgroid gamerule, watch vgroid still be there

## Media
image

image

image

**Changelog**

:cl:
- add: Banners (the ones you can craft) can now be used to claim debris/vgroid/UIV/etc grids via altclick if anchored, preventing them from despawning.


Ilya246:
adding guidebook entry, don't merge before next commit

Ilya246:
done, ready

Cojoke-dot:
Should be tied to ids like ship deeds I think

Ilya246:
> Should be tied to ids like ship deeds I think

tied to what ids and how

Cojoke-dot:
I guess for the future of the system. Claiming rocks should be limited to one per id for lag, like ship deeds

Ilya246:
> I guess in the future I guess. Claiming rocks should be limited to one per id for lag, like ship deeds

as i said, if people claim too many, which i don't think will happen, then limit it
what i wanted to do in case people end up overusing it was to just make it use a special banner and have a limited amount sold at some POI

Redrover1760:
It'd be cool if grid claiming gave the grid a company thing too that'd show up on IFF. Which would presumably be based off the ID just like ship buying does.

Ilya246:
> It'd be cool if grid claiming gave the grid a company thing too that'd show up on IFF. Which would presumably be based off the ID just like ship buying does.

i think we could just have 2 separate types of claiming, "claiming" like here which just prevents despawn so you don't have to just sit near the grid so it doesn't despawn, and proper claiming where you take ownership which sets iff and lets you name it and all
maybe second one lets you install crew records console and such on it

OneZerooo0 (approved):

PR #2005, opened 2025-08-23
Vgroid Deep Gas Deposits

Ilya246:
## About the PR
vgroids now have deep gas deposits
they work like factorio oil, have yield which decreases based on base extraction rate
yield can never drop below a minimum so you can have infinite gas at a lower rate
also adds rare deep frezon deposits to vgroids only

## Why / Balance
more incentive to colonise vgroid
atmos colony gameplay that doesn't need gas miners
note that considering vgroids are large those will generally need long pipelines to actually get gas to your ship
makes gas deposit scanner actually useful
frezon deposits are probably fine since i gave them a reduced chance to appear and it doesn't yield that much frezon anyway

## How to test
`addgamerule BluespaceDungeonSnow`
fly to it
go around with gas deposit scanner
find deposit, mine it

## Media
image

image

since screenshots were taken the `remains .` part was fixed

**Changelog**

:cl:
- add: VGroids now have deep gas deposits, which can be gas mined indefinitely.

### DIFF
--- a/Content.Server/_NF/Atmos/Components/RandomGasDepositComponent.cs
+++ b/Content.Server/_NF/Atmos/Components/RandomGasDepositComponent.cs
@@ -17,6 +17,7 @@ public sealed partial class RandomGasDepositComponent : Component
     /// <summary>
     /// A scale factor on the deposit's size.
     /// After each gas is chosen from the deposit prototype, the scale factor is multiplied into the deposit size.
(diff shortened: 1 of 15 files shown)


Avalon-Proto:
Is the snow Roid the only one with gas deposits?

Ilya246:
> Is the snow Roid the only one with gas deposits?

all vgroids have gas deposits, but snow has more and scrap has less but gets a guaranteed frezon deposit

PR #2085, opened 2025-08-27
Revert Dewreckening

Ilya246:
## About the PR
reverts #2010

## Why / Balance
this made wrecks very very uncommon
i don't think they even lag anyway, they don't actually even generate anything on themselves until you get 16m within them
even then they despawn once you fly away

## How to test
fly around
see that wrecks actually exist

**Changelog**

:cl:
- tweak: Wrecks should be common again.


Redrover1760 (approved):
Sounds about right. No clue the source of the lag at this point.

PR #2091, opened 2025-08-27
Make Large Thrusters Not Terrible

Ilya246:
## About the PR
changed from taking as much power as 6.7 normal thrusters and providing as much thrust as 2.3 to providing as much thrust as 4 AND providing bonus max speed equal to +50% speed machine part upgrades

## Why / Balance
it's kinda terrible right now and there's no reason to not just replace it with 4 normals
however with new stats being actually usable the amount capitals currently have might be a bit overtuned and need changing
we can try it and see

## How to test
get anything with large thrusters
go forward
whee

## Media
tested, works

**Changelog**

:cl:
- tweak: Large thrusters now provide more acceleration and provide bonus max speed.


Redrover1760:
Testfail is unrelated heisenfail :(

If your branch is updated anyways

Ilya246:
merged main, no more testfail

PR #2092, opened 2025-08-27
Obscure IFF Function (WIP)

OneZerooo0:
WIP
## About the PR

Adds a new function for the IFF console, for obscuring IFF details - E.g rather than either "Serpent TSF-420" or "Olympus CIV-690", its "corvette" or "cruiser" or something more vague. Rogues will probably keep the full IFF hiding capability.

When finished, almost all IFF consoles will no longer be able to completely hide IFF, only hide incriminating details. This would take the form of replacing signature with a generic term shared by other ships in the obscured ship’s weight class.

## Why / Balance

Cool. Hopefully with this change, ships are harder to completely disappear (particularly when leaving combat.) 
## How to test

Turn on Obscure IFF, go off the ship and look at a mass scanner.
## Media

not yet
## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Non-rogue IFF consoles now obscure IFF details rather than completely hide them.

### DIFF
--- a/Content.Client/Shuttles/UI/IFFConsoleWindow.xaml.cs
+++ b/Content.Client/Shuttles/UI/IFFConsoleWindow.xaml.cs
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: 2022 metalgearsloth
 // SPDX-FileCopyrightText: 2022 wrexbe
 // SPDX-FileCopyrightText: 2025 Onezero0
+// SPDX-FileCopyrightText: 2025 significant harassment
+// SPDX-FileCopyrightText: 2025 sleepyyapril
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
--- a/Content.Client/Shuttles/UI/IFFConsoleWindow.xaml.cs
+++ b/Content.Client/Shuttles/UI/IFFConsoleWindow.xaml.cs
@@ -1,3 +1,9 @@
+// SPDX-FileCopyrightText: 2022 metalgearsloth
+// SPDX-FileCopyrightText: 2022 wrexbe
+// SPDX-FileCopyrightText: 2025 Onezero0
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
 using Content.Client.Computer;
 using Content.Client.UserInterface.Controls;
 using Content.Shared.Shuttles.BUIStates;
--- a/Content.Client/Shuttles/UI/IFFConsoleWindow.xaml.cs
+++ b/Content.Client/Shuttles/UI/IFFConsoleWindow.xaml.cs
@@ -14,9 +14,10 @@ public sealed partial class IFFConsoleWindow : FancyWindow,
 {
     private readonly ButtonGroup _showIFFButtonGroup = new();
     private readonly ButtonGroup _showVesselButtonGroup = new();
+    private readonly ButtonGroup _obscureIFFButtonGroup = new(); // _Mono
     public event Action<bool>? ShowIFF;
     public event Action<bool>? ShowVessel;
-
+    public event Action<bool>? ObscureIFF; // _Mono
     public IFFConsoleWindow()
     {
         RobustXamlLoader.Load(this);
@@ -29,6 +30,13 @@ public sealed partial class IFFConsoleWindow : FancyWindow,
         ShowVesselOnButton.Group = _showVesselButtonGroup;
         ShowVesselOnButton.OnPressed += args => ShowVesselPressed(true);
         ShowVesselOffButton.OnPressed += args => ShowVesselPressed(false);
+
+        // _Mono: Obscure button groups & OnPressed(s)
+        ObscureIFFOffButton.Group = _obscureIFFButtonGroup;
+        ObscureIFFOnButton.Group = _obscureIFFButtonGroup;
+        ObscureIFFOnButton.OnPressed += args => ObscureIFFPressed(true);
+        ObscureIFFOffButton.OnPressed += args => ObscureIFFPressed(false);
+        // End mono
     }
 
     private void ShowIFFPressed(bool pressed)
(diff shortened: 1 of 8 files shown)


OneZerooo0:
Opened for tests

Content.Client/Shuttles/UI/IFFConsoleWindow.xaml.cs
```diff
@@ -14,9 +20,10 @@ public sealed partial class IFFConsoleWindow : FancyWindow,
 {
     private readonly ButtonGroup _showIFFButtonGroup = new();
     private readonly ButtonGroup _showVesselButtonGroup = new();
+    private readonly ButtonGroup _obscureIFFButtonGroup = new(); // _Mono
     public event Action<bool>? ShowIFF;
     public event Action<bool>? ShowVessel;
-
+    public event Action<bool>? ObscureIFF; // _Mono
```
sleepyyapril, line 29:
```suggestion
    public event Action<bool>? ObscureIFF; // _Mono

```

Content.Client/Shuttles/UI/IFFConsoleWindow.xaml.cs
```diff
@@ -14,9 +20,10 @@ public sealed partial class IFFConsoleWindow : FancyWindow,
 {
     private readonly ButtonGroup _showIFFButtonGroup = new();
     private readonly ButtonGroup _showVesselButtonGroup = new();
+    private readonly ButtonGroup _obscureIFFButtonGroup = new(); // _Mono
```
sleepyyapril, line 25:
```suggestion
    private readonly ButtonGroup _obscureIFFButtonGroup = new(); // _Mono

```

sleepyyapril (changes requested):
spacing matters

Redrover1760 (changes requested):
Can you describe to me the end goal behind this change?

As far as I'm aware, all this will do is make it stupid easy to avoid TSFMC and USSP while allowing rogues to do whatever they want with actual no IFF ships.

We already have issues with it being absurdly easy to run away and/or FTL away from ships.

You should also make very clear what the end result of your changes are in you PR body.

starch70:
I think this would be better to just replace normal IFF label hide anyways (for ships, not POIs), atleast for large FTL map
Cuck consoles to hide you from everything are boring and i want to find ships to frag

Redrover1760:
> I think this would be better to just replace normal IFF label hide anyways (for ships, not POIs), atleast for large FTL map Cuck consoles to hide you from everything are boring and i want to find ships to frag

would require some code that only show obscured ships in an area at the very least, else one can't have secret bases/locations.

starch70:
> > I think this would be better to just replace normal IFF label hide anyways (for ships, not POIs), atleast for large FTL map Cuck consoles to hide you from everything are boring and i want to find ships to frag
> 
> would require some code that only show obscured ships in an area at the very least, else one can't have secret bases/locations.

Stealthkiller code for spear reveals fullcloak ships within a radius, you could probably just edit it easily to just IFFcloak ships in a radius instead.

OneZerooo0:
True

OneZerooo0:
> Can you describe to me the end goal behind this change?
> 
> As far as I'm aware, all this will do is make it stupid easy to avoid TSFMC and USSP while allowing rogues to do whatever they want with actual no IFF ships.
> 
> We already have issues with it being absurdly easy to run away and/or FTL away from ships.
> 
> You should also make very clear what the end result of your changes are in you PR body.

I believe this would help with the wild goose chasing of IFFless ships - right now if you have an IFF console, you can just.. leave, and never be found again unless you want to be. The reason I’m not sure about Rogues is because a lot of their gameplay revolves around stealth, so removing that would probably break a lot of stuff. Starch’s idea might be good though
Updating PR body now

Content.Server/Shuttles/Systems/ShuttleSystem.IFF.cs
```diff
@@ -103,6 +112,26 @@ private void OnIFFShowVessel(EntityUid uid, IFFConsoleComponent component, IFFSh
         }
     }
 
+    // _Mono: Obscure IFF handling
+    private void OnIFFObscureVessel(EntityUid uid, IFFConsoleComponent component, IFFObscureIFFMessage args)
+    {
+        if (!TryComp(uid, out TransformComponent? xform) || xform.GridUid == null ||
```
Cojoke-dot, line 118: Don't use TryComp for TransformComponent, use:
var xform = Transform(uid);

Redrover1760:
> > Can you describe to me the end goal behind this change?
> > As far as I'm aware, all this will do is make it stupid easy to avoid TSFMC and USSP while allowing rogues to do whatever they want with actual no IFF ships.
> > We already have issues with it being absurdly easy to run away and/or FTL away from ships.
> > You should also make very clear what the end result of your changes are in you PR body.
> 
> I believe this would help with the wild goose chasing of IFFless ships - right now if you have an IFF console, you can just.. leave, and never be found again unless you want to be. The reason I’m not sure about Rogues is because a lot of their gameplay revolves around stealth, so removing that would probably break a lot of stuff. Starch’s idea might be good though Updating PR body now

so would removing the ability for TSF and USSP to hide in order to try and hunt down rogues. If I were a rogue I'd just warp away any time a "Cruiser" or "Destroyer" got within 1000m of me.

Content.Shared/Shuttles/Systems/SharedShuttleSystem.IFF.cs
```diff
@@ -71,13 +83,17 @@ public Color GetIFFColor(EntityUid gridUid, bool self = false, IFFComponent? com
...
             // Return a formatted label that the client can parse properly
             return $"{labelText}\n{companyName}";
         }
+        else
+        {
+            return $"{labelText}";
+        }
 
-        return labelText;
+        return $"{vesselClass}";
```
Cojoke-dot, line 96: Will this return ever run? It's right after an else statement that also has a return

Content.Shared/Shuttles/Components/IFFComponent.cs
```diff
@@ -51,6 +69,12 @@ public enum IFFFlags : byte
     /// </summary>
     Hide = 2,
 
+
+    /// <summary>
+    /// Frontier - Is this a player shuttle
```
Cojoke-dot, line 74: don't forget to fix this, we ain't in frontier

Cojoke-dot (changes requested):
Couple small things, some are just questions. idk what ide you are using but don't ignore the yellow lines if you got a warning for that TryComp<TransformComponent>

OneZerooo0:
closing for now since initial draft of this isn't following a new goal + ilya is probably reworking IFF anyway + bloating PR count

PR #2103, opened 2025-08-28
Borg Buffs

Ilya246:
## About the PR
- borgs now get inbuilt mass scanner
- borgs now get MovementIgnoreTouching
- "appraisal module" -> "cargo module" + tethergun + gravity gun
- all borg types now get tool module (all the borgs just get it installed anyway already)
- medical borg gets surgery module
- borgs can now open their own UI
- you can now eject default modules

## Why / Balance
lets borgs not be helpless in space
makes them useful to have for space operations if you don't have voidboots since they can just fly
appraisal module was useless so now isn't, tethergun isn't even that strong
also currently tethergun/gravgun are missing from research tree because frontier so should maybe be added in another PR
as for the last 3 changes, why was it not the case already

## How to test
obvious

## Media
image

image

**Changelog**

:cl:
- add: Borgs now have inbuilt mass scanner and can freely float in space.
- tweak: Borg appraisal module renamed to cargo module and given tether gun and gravity gun.
- tweak: Medical borgs now get the surgery module right away.
- tweak: All borgs now get the tools module right away.
- tweak: Borgs can now open their own UI.
- tweak: You can now eject default modules from borgs.

### DIFF
--- a/Content.Client/Silicons/Borgs/BorgMenu.xaml.cs
+++ b/Content.Client/Silicons/Borgs/BorgMenu.xaml.cs
@@ -1,3 +1,12 @@
+// SPDX-FileCopyrightText: 2024 Nemanja
+// SPDX-FileCopyrightText: 2024 Pieter-Jan Briers
+// SPDX-FileCopyrightText: 2024 ShadowCommander
+// SPDX-FileCopyrightText: 2024 metalgearsloth
+// SPDX-FileCopyrightText: 2025 Ilya246
+// SPDX-FileCopyrightText: 2025 Whatstone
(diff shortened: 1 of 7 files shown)


Redrover1760:
you could give all borg types innate jetpack modules although even that is annoying because you can only fly with it active and can't do anything else

Ilya246:
> you could give all borg types innate jetpack modules although even that is annoying because you can only fly with it active and can't do anything else

i think this is better because yeah that, and so you need module to fly fast
the 1-item modules in general could all just use some buffs

PR #2123, opened 2025-08-29
Emp Arcs Real

Ilya246:
## About the PR
makes it possible for bullets to have emp arcs like starsector

## How to test
add it to autopulser the way described in the yml comments
fire autopulser at ship
see emp arc

## Media
demonstration video of it being added to autopulser

https://github.com/user-attachments/assets/204af00f-deb8-4388-9961-b8e3bc4a32cc


Ilya246:
> Just give the EMP effect to dedicated EMP shipguns and/or make the range and duration/energy drain less strong for the autopulser one

is now just the code and not the autopulser change

PR #2130, opened 2025-08-30
Fix Rapid Borg Cell Drain

### LINKED ISSUE #2128: Cyborg Cells drain VERY rapidly upon powering on.
## Description
For some reason, when a Cyborg enters its chassis, its cell drains VERY quickly, even Microreactors drain in ~20 seconds.

**Reproduction**
- Give a Positronic + Battery to a Cyborg.
- Enter the ghost role.
- 100 to 0 in not even a minute.
- Pain.

**Additional context**
Seems to fix itself randomly, don't know what causes it to stop.

Ilya246:
## About the PR
fixes #2128
this is why i prefer accumulators
this really should be ported to wizden though

## How to test
become borg after long time of it being off
see power not draining rapidly unlike it would've otherwise

**Changelog**

:cl:
- fix: Fixed borg power cells draining rapidly after turning on after a long time of being off.


Fat-Engineer-Gaming:
Thank you...

PR #2132, opened 2025-08-30
Remove Hardcoded Lathe Slowdown

Ilya246:
## About the PR
apparently frontier hardcoded an untagged 3x lathe slowdown and this is why they are so slow

## Why / Balance
lathes are currently painfully slow anyway so this is good
lathes being fast doesn't hurt anyone

## How to test
use lathe
no longer 3x slower than it should be

**Changelog**

:cl:
- tweak: Lathes are no longer painfully slow (sped up 3x).


tonotom1:
another certified frontier method for balancing around 12 hour shifts

burrrkeye:
frontier would be that way 😡👉

OneZerooo0 (approved):

OneZerooo0:
hardcode…

Redrover1760:
Wtf frontier

PR #2210, opened 2025-09-03
Get Rid of Voucher Accesses

Ilya246:
## About the PR
vouchers can now just specify a list of vessels they can be used for

## Media
image

**Changelog**

- tweak: Shipyard console now tells you when a ship is voucher-only.


Ilya246:
> all the ships that were only accessible via voucher before (like flyssa and ledokol (i think?)) should be CC access too

i'll just add purchasable bool

Ilya246:
image

starch70 (approved):

PR #2213, opened 2025-09-03
AI/Borg Understands All Languages

Honestly101:
## About the PR
Now they can serve as a translator due to their extensive database of known languages.

**Changelog**
:cl:
- tweak: AI/Cyborgs can now speak/understand all languages


Cojoke-dot:
Meh? They shouldn't have faction languages I think. Kinda beats the whole encryption through language thing if my ship ai can instantly break it. 

Doesn't make much sense with the posi and mmi brains. Why would ipc and any other species learn every language after leaving their body.

Fix your indentation for the MMI...

PR #2249, opened 2025-09-07
Hoverbike Changes

ScyronX:
## About the PR

This PR makes hoverbikes better in space but worse on land/gravity.
## Why / Balance

Hoverbikes were too unfair in combat when it came to ground combat. I believe certain engine changes made the movement speed modifier component behave differently leading to a different mobility then originally intended, so I fixed it while improving and giving it more contrast between space and land.

## Requirements

**Changelog**

:cl:
- tweak: Hoverbikes are worse on land and better in space.


Ilya246:
add `base` before every stat (like `baseAcceleration`) or it won't work

ScyronX:
> add `base` before every stat (like `baseAcceleration`) or it won't work

Yes, thank you, I almost forgot to change one of the components. I believe this was the problem with the older component values . It was not setting base values so the mobility was way better on ground then it should have been.

Redrover1760 (approved):

PR #2289, opened 2025-09-12
Chimera Change

Ilya246:
## About the PR
- refactors chimera yml a bit
- chimera faction changed to attack everything
- chimeras will now attack even crit mobs
- infecting a non-species mob now only produces a weaker flesh monster and not full fleshbeast
- changed chimeras to not bleed out for some reason
- removes juking so they just run at you

## Why / Balance
500 chimeras from one monkey box too easy
you still get beasts though and they aren't weak either

## Media
image

also tested with every species, properly turns into normal chimera

**Changelog**

:cl:
- tweak: Chimera AI changed to also attack crit mobs.
- tweak: Infecting non-species mobs with letoferol now only makes a weaker flesh monster and not a full chimera.

### DIFF
--- a/Resources/Prototypes/_Mono/Entities/Mobs/Chimera/htn.yml
+++ b/Resources/Prototypes/_Mono/Entities/Mobs/Chimera/htn.yml
@@ -1,3 +1,4 @@
+# SPDX-FileCopyrightText: 2025 Ilya246
 # SPDX-FileCopyrightText: 2025 starch
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
--- a/Resources/Prototypes/_Mono/Entities/Mobs/Chimera/htn.yml
+++ b/Resources/Prototypes/_Mono/Entities/Mobs/Chimera/htn.yml
@@ -70,9 +70,6 @@
             pathfindKey: TargetPathfind
             rangeKey: MeleeRange
         - !type:HTNPrimitiveTask
-          operator: !type:JukeOperator
-            jukeType: Away
-        - !type:HTNPrimitiveTask
           operator: !type:MeleeOperator
             targetKey: Target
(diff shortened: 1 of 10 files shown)


Resources/Prototypes/_Mono/Entities/Mobs/Chimera/htn.yml
```diff
@@ -66,4 +50,54 @@
       task: EscapeCompound
   - tasks:
     - !type:HTNCompoundTask
-      task: MeleeAttackTargetCompound
+      task: ChimeraMeleeAttackTargetCompound
+
+# MeleeAttackTargetCompound but also attacks dead and crit
```
Redrover1760: could this result in a chimera infinitely attacking something that can't turn (like an IPC that doesn't have chems)

Avalon-Proto: I mean, it's possible. But I think it would retarget to a living thing that approaches it still

Ilya246: > could this result in a chimera infinitely attacking something that can't turn (like an IPC that doesn't have chems)

feature

Ilya246: actually just tested and they do not attack dead IPC

Ilya246: or no they sometimes do sometimes don't it's behaving weirdly

Ilya246: changed it to attack only alive and crit mobs

Redrover1760 (approved):

PR #2304, opened 2025-09-13
Buffs the Autopulsar and L85 20mm

Redrover1760:
## About the PR

L85 gains structural damage, I apparently failed to buff it the first time and buffed the wrong ammo type instead :(

Gives the autopulsar even more structural damage

## Why / Balance

this thing does no damage despite having the downside of running out of ammo and having to recharge rather quickly.

other is just a fix for a failed buff attempt of 20mm

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Autopulsar damage buff


OneZerooo0 (approved):
both could actually be buffed more but this is at least better than before

Redrover1760:
> both could actually be buffed more but this is at least better than before

True I'll look at maybe buffing more to be fair both have very high rates of fire though and you often have quite a lot on a ship

Redrover1760:
I love bullying children

PR #2334, opened 2025-09-20
Hammerhead, Merc Mil Ship

OneZerooo0:
## About the PR

Replaces the ZOB Claymore with the SHM Hammerhead (now in the exped shipyard), a similar frontal destroyer armed with two TARNYX, two Cyrexa, and four dravon. In exchange for better sightlines, multiple crew is recommended for piloting to reload, plus the fact that there is less armor overall. Also cheaper. Also also, it can do expeds.
## Why / Balance

Claymore was slop - interior and exterior was boring, it was too much of an ez brainless frag ship, and for some reason it was in the civ shipyard.

## How to test

buy hammerhead, look to see it + claymore is atomized
## Media

image
image

## Requirements

## Breaking changes

**Changelog**

:cl: Onezero0
- add: Added the SHM Hammerhead to the expedition shipyard.
- remove: Removed the ZOB claymore.


Ed-Head:
I understand the idea was to make a ship that is less 'slop' and part of that was some clear weak-points but I honestly think a few too many have been added to the point of making the whole design somewhat questionable for use in combat. 

The most glaring point of course being the entrance seemingly specifically for boarders located at the back of the ship, it leads to the AME and no-where else. said hallway cant even be accessed from where the crew will be during combat. Am I supposed to hire a guy to hang out there or just wall it off?

The only point of that back entrance seems to be to let boarders jetpack up uncontested (because the 20mm guns are at the front) and put c4 on their AME. Why?

The other questionable design is the obvious weak-point in its frontal armor system; the dock. If the armor is to be understood then most fire this ship expects to take is frontal. Why then is there an airlock that cant take much abuse at the front leading to an open hallway that connects to the bridge?

Using the side airlocks as the docking ports would have made much more sense for this and would not undermine its role as a frontal destroyer.

These obvious weaknesses wouldn't be as painfully glaring if the overall armor wasn't so thin. Beyond that, the use of armor is utterly puzzling to me. Why put plast armor behind the dravons and the empty space above them instead of the bridge? Or behind the engine? 

I'm not saying make the thing invincible, but this is absurd. If the idea was supposed to be glass cannon I think you added too much glass.

Rubeebeebee:
aw man theyre removing the claymore :(

burrrkeye:
are we going for non-existent armor plating on merc ships, or??

also love the ragebait rear facing gun console always a nice troll

OneZerooo0:
feedback recieved, my intent was not to make it too much of a glass cannon

Rubeebeebee:
also, the claymore was in the CIV console because it was a non-exped capable merc ship, like the sellsword.

Ilya246:
claymore is good but otherwise yeah this isn't too usable, merc ships are already weak and claymore was one of the usable ones
also what's even the point of arc-locked 90mm on a ship this size, it won't ever get used
why would you ever orient your slow-rotating ship to use the 90mm on this when the larger caliber guns exist
it could be improved if you gave it more gyros and thrusters at least
people will probably just cut off the side things on the larger caliber guns to give them a 180 firing arc and move the 90mm forward to give them firing arcs, might as well have that by default

i think that if there's something you would want to immediately refit at low cost upon buying a ship then that ship is badly designed and this ship has several of this

Skyguard12:
Combat ship with reinforced walls. Always funny.

PR #2347, opened 2025-09-24
No Gas Purity Penalty At Edison

bitcrushing:
## About the PR

Adds a GasPriceNoPurity method to AtmosphereSystem.Utils.cs to be used by Edison gas sale consoles, suggested by people on the Discord server.

## Why / Balance

Having to wait 10mins for a gas to clear out of the Edison pipenet so you can sell it without a penalty is not fun
Makes gas mining purely for sales more viable

## How to test

Pipe mixed and unmixed gases into a gas sale port
Use gas sale console to look at price, compare mixed and unmixed to see they add up

## Requirements

## Breaking changes

None, alternative method to GasPrice

**Changelog**

- tweak: No more gas purity malus on mixed gas sales at Edison

### DIFF
--- a/Content.Server/Atmos/EntitySystems/AtmosphereSystem.Utils.cs
+++ b/Content.Server/Atmos/EntitySystems/AtmosphereSystem.Utils.cs
@@ -28,11 +28,26 @@ public partial class AtmosphereSystem
 
         // Pay more for gas canisters that are more pure
         float purity = 1;
-        if (totalMoles > 0) {
+        if (totalMoles > 0)
+        {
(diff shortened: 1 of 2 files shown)


Ilya246:
why add whole new method for this just make it a bool in the original

bitcrushing:
> why add whole new method for this just make it a bool in the original

don't wanna shit up upstream code too much since GetPrice is used to measure normal canister prices as well

i could move it into GasDepositSystem or it's own thing under _Mono now that i think about it since it's not used anywhere else, but having two very similar methods split across files like that would feel awkward

open to suggestions

PR #2350, opened 2025-09-25
Magus Update

Ghagliiarghii:
## About the PR

The Magus is a med ship so I wanted to give it some of the basics a lot of med ships come with to make medical gameplay in this ship better.

To that end:
-added crew monitoring console to cockpit to locate dead, crit, or injured civilians to go rescue
-added withdrawl only atm in airlock so those rescued civilians can (optionally) pay for that treatment
-added sink to chemistry area so a chemist can get water from a less limited source than the water cooler in the bridge
 (These are all wallmounted and so cannot be removed from the ship to the best of my knowledge)
 
Additionally:
-removed two small corner railings one of which made it impossible to unanchor and then reanchor the merc fab and the other for symmetry
-moved the shield generator to the mercenary side and the FTL drive to the medical side since I thought they would fit better thematically there
-added a couple more HVs underneath the AME for no reason (makes it slightly easier to swap the generator with RTGs in place should you be so fortunate as to find enough)

-Increased price a little since, while appraisegrid gives me ~40,000 value in the mapping dimension it's closer to 60,000 after I purchase it due to children of filled lockers etc. 
## Media

image
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Slight Magus QoL improvements to make it a more capable medical vessel. Price raised by 9,000 credits to reflect this.

### DIFF
--- a/Resources/Maps/_Mono/Shuttles/Expedition/magus.yml
+++ b/Resources/Maps/_Mono/Shuttles/Expedition/magus.yml
@@ -22,8 +22,8 @@ meta:
   engineVersion: 264.0.0
   forkId: ""
   forkVersion: ""
-  time: 08/10/2025 13:43:51
-  entityCount: 653
+  time: 09/25/2025 16:22:11
+  entityCount: 661
 maps: []
 grids:
 - 1
@@ -414,7 +414,7 @@ entities:
           2,-1:
             1: 8705
           0,4:
-            1: 16
+            1: 528
         uniqueMixes:
         - volume: 2500
           temperature: 293.15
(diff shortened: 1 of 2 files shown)


Ghagliiarghii:
Also this ship was created by @OneZerooo0 so I would prefer to have their approval for these changes before this is merged if possible.

OneZerooo0 (approved):
magus still not overpowered trust
-not a biased maintainer
Price looks nice, all clear there
Thanks for your work :)

Danchi299:
Are we treating the missing tile under the Expeditionary Console as intended now?

PR #2374, opened 2025-10-03
Vulture Laser Armament 

Arcticular1:
## About the PR

Replaced Bofors on Vulture with L-20 Apollo Heavy Laser Cannons and upgrades the gunnery server to accommodate for this. 
## Why / Balance
The Vulture is very ineffective in combat and the main battery is very ineffective at mining, usually destroying every bit of ore in it's path. 

This change allows the vulture to effectively take out targets that are infront of it. It's firepower is much worse than similar capitals of it's class so it will not be winning any battles. However it can effectively damage targets that enter the range of it's main gun battery.

The only issue that i can see with this change is the fact that i put a higher power gunnery server on the vulture. This can be traded off for taking away some 90mms to compensate for the lasers.

(Also its a planetcracker and those are usually lasers!!! pew pew!!)
## How to test

Spawn it in at the blackmarket shipyard and try it out.
## Media

Pretty good at destroying ore without deleting it.
image
image

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Maps/_Mono/Shuttles/BlackMarket/vulture.yml
+++ b/Resources/Maps/_Mono/Shuttles/BlackMarket/vulture.yml
@@ -1,3 +1,23 @@
+# SPDX-FileCopyrightText: 2023 Leon Friedrich
+# SPDX-FileCopyrightText: 2023 Nemanja
+# SPDX-FileCopyrightText: 2023 Scribbles0
+# SPDX-FileCopyrightText: 2023 TsjipTsjip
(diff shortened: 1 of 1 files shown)


Arcticular1:
Forgot to mention you cant unanchor the Apollos to abuse it :)

graycatsaderow:
are they better or worse at destroying ships

Arcticular1:
> are they better or worse at destroying ships

Quite a bit better

OneZerooo0 (approved):
screw it its funny

PR #2336, opened 2025-09-21
Reapply "Port Goob Borg Stuff (#2104)"

Ilya246:
## About the PR
revert didn't fix borglag

**Changelog**

:cl:
- tweak: Returned borg sprite changing because it wasn't the cause of borg lag.

Issue #2224, opened 2025-09-05
Cyborgs just fucking kill everything


Fat-Engineer-Gaming:
## Description
When near a borg, you start experiencing rubberbanding.

**Reproduction**
Be near a borg, it will start rubberbanding after a bit.

thoughtlessuser:
i second this

VERY known issue

pls fix soon 

its actually painful

WAYS TO FIX: 
Clear engine and cache in the hub and rejoin 
(It only works briefly ):)

Redrover1760:
@Ilya246

Ilya246:
i have no idea

Fat-Engineer-Gaming:
i blame slopstice for literally no reason

Ilya246:
@SolsticeOfTheWinter why did you do this

burrrkeye:
I magdump clankers wordlessly because of this, they can just respawn as a real player instead

Danchi299:
It constantly threw a client-side error in client console about invalid person speaking, try screenshoting console next time someone lags due to a borg.

PR #2477, opened 2025-10-25
Borg Adjustments

EckoAurum:
## About the PR
In the Sisyphean challenge of fixing the localized borg rubberbanding lag #2224, players have identified that cyborg lights have _something_ to do with it, possibly.

This cleans up some of the issues.
**First:** Removed ALL lights from cyborgs in their entirety. Added blue tinted night vision in it's place.

**Second:** Re-adds borg sprites. This was identified as not the issue in #2336, but they're currently still broken. I'm not entirely sure why it works on other forks, but not ours. Adding the SwitchableSubtype comp on initialization seems to fix it for now.

**Third:** Gave the medical cyborg a starting Defibrillator. I intended to fix Medical cyborg not being able to do surgery through clothes, since they cannot strip, but our shitmed is behind and that component doesn't exist without updating our shitmed.

## Why / Balance
**First:** Lights are the possible issue. If they aren't, we can look elsewhere. Nightvision also just fits as it's become quite commonplace.

**Second:** It's nice being able to tell what borg is what,

**Third:** Medical cyborgs can't do surgery without someone to help them. Though we can't fix that yet, In Mono most people are dead by the time you can actually tend to them. Giving them a defib at the start allows them to do their job effectively.

## How to test
Spawn in Cyborgs and play around.

## Media
image

Blue Tint since removed.

image

## Requirements

## Breaking changes
None found so far.

**Changelog**

:cl:
- add: Medical Cyborgs now start with the defibrillator module.
- remove: Removed all lights from Cyborgs.
- add: Cyborgs now have Nightvision.
- tweak: Returned Cyborg sprites.

### DIFF
--- a/Content.Shared/Silicons/Borgs/SharedBorgSwitchableTypeSystem.cs
+++ b/Content.Shared/Silicons/Borgs/SharedBorgSwitchableTypeSystem.cs
@@ -52,6 +52,7 @@ public abstract class SharedBorgSwitchableTypeSystem : EntitySystem
     private void OnMapInit(Entity<BorgSwitchableTypeComponent> ent, ref MapInitEvent args)
     {
         _actionsSystem.AddAction(ent, ref ent.Comp.SelectTypeAction, ActionId);
+        EnsureComp<BorgSwitchableSubtypeComponent>(ent.Owner); // Mono - Temp fix for borg sprites
         Dirty(ent);
 
         if (ent.Comp.SelectedBorgType != null &&
--- a/Content.Shared/Silicons/Borgs/SharedBorgSwitchableTypeSystem.cs
+++ b/Content.Shared/Silicons/Borgs/SharedBorgSwitchableTypeSystem.cs
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: 2024 Pieter-Jan Briers
 // SPDX-FileCopyrightText: 2025 BeBright
+// SPDX-FileCopyrightText: 2025 EckoAurum
 // SPDX-FileCopyrightText: 2025 Ilya246
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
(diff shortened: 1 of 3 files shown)


Avalon-Proto:
I would standardize the NVG w/ the same colors as the rest. Blue can be hard on the colorblind, so use the same shader as all the other NVG options

EckoAurum:
> I would standardize the NVG w/ the same colors as the rest. Blue can be hard on the colorblind, so use the same shader as all the other NVG options

Was debating between the two and saw Asakim's NVG's werent changed and yoinked their blue. I'll set it back to default!

Ilya246:
adding a comp on mapinit fails tests, just add it to the proto if it's not there, otherwise i don't even know

EckoAurum:
You can see that all the checks passed in commit "Comment out instead of delete" bc9e1e422ff4f4b34a1772ebb8fee0c44043e88e

There's no way changing the color of the night vision causes the failure of a DeviceNetwork entity on sevastopol, and causing a ship to cost 56000 instead of 56024.113688498735 on only Ubuntu. It's not real, it can't be real.

Ilya246:
> You can see that all the checks passed in commit "Comment out instead of delete" [bc9e1e4](https://github.com/Monolith-Station/Monolith/commit/bc9e1e422ff4f4b34a1772ebb8fee0c44043e88e)
> 
> There's no way changing the color of the night vision causes the failure of a DeviceNetwork entity on sevastopol, and causing a ship to cost 56000 instead of 56024.113688498735 on only Ubuntu. It's not real, it can't be real.

strange, then it probably already has the component but ensuring it anyway somehow fixes it?
also, there's no "only ubuntu" because there are only ubuntu tests, nothing to be "only" against

and this testfail isn't your fault, yes

ark1368 (approved):

PR #2514, opened 2025-10-29
Infinite Shipgun Backup Ammo

Ilya246:
## About the PR
makes shipguns use an infinite pool of backup ammo when not specifically loaded
likely jank due to the way guncode works but fully worked in testing
makes gunsole show "name: ammo" instead of just ammo when gun selected
DNM until @kasature90 review
may need us to add "standard" ammo types for this to use

## Why / Balance
requested and good

## How to test
fire cyrexa/20mm/whatever without ammo
works
load special ammo
see it fire special ammo instead then switch back

## Media
image
image

**Changelog**

:cl:
- tweak: Shipguns now have an infinite pool of their normal ammo, but can still be loaded to fire special ammo.


kasature90 (changes requested):
Should use worse-than-standard ammo (or implement special ammo types)
If making the missiles infinite ammo, 1/3 their special ammo count (currently its just 3x the amount of visible tubes)

Info about this should probably be put in guidebook too so it isnt confusing

Ilya246:
> Should use worse-than-standard ammo (or implement special ammo types) If making the missiles infinite ammo, 1/3 their special ammo count (currently its just 3x the amount of visible tubes)
> 
> Info about this should probably be put in guidebook too so it isnt confusing

would reusing the pre-ammo-update projectiles be ok or did it not change projectiles

kasature90:
> would reusing the pre-ammo-update projectiles be ok or did it not change projectiles

it did not change projectiles

Skyguard12:
Hold on... what's even the point of ammo if it's infinite anyway?

Ilya246:
> Hold on... what's even the point of ammo if it's infinite anyway?

you can load special ammo to get more out of your shipguins

Ilya246:
> Should use worse-than-standard ammo (or implement special ammo types) If making the missiles infinite ammo, 1/3 their special ammo count (currently its just 3x the amount of visible tubes)
> 
> Info about this should probably be put in guidebook too so it isnt confusing

made the default-infinite ammo have 80% the stats of normal ammo
also a bunch of visual/examine fixes

kasature90:
> made the default-infinite ammo have 80% the stats of normal ammo

tbh i doubt that only 25% better will be enough to get people to do reloading

Ilya246:
> > made the default-infinite ammo have 80% the stats of normal ammo
> 
> tbh i doubt that only 25% better will be enough to get people to do reloading

intent is we buff the special ammo later
and i think it's fine to try for now

Ilya246:
BILLIONS must merge master

Ilya246:
ok i looked and this should be a master testfail but somehow fails for my new proto specifically (???)

ark1368 (approved):
Ok

PR #2537, opened 2025-11-01
New Mixed Gamemode

Ilya246:
## About the PR
- adds new gamemode which has both ADS and chimera schedulers but 1.5x slower
- fixes mode descriptions to not say anything about USSP anymore

## Why / Balance
ADS has interactions with chimera and variety is just good

**Changelog**

:cl:
- add: Added new mixed gamemode with both ADS and chimera.


Avalon-Proto:
Honestly I'd rather only have this instead of the 2 modes, but a bit faster

Ilya246:
> Honestly I'd rather only have this instead of the 2 modes, but a bit faster

we can try it out and disable the other ones later or speed this one up if we want

Skyguard12:
I just want Apoc back, man. Apoc was peak.

OneZerooo0 (approved):

OneZerooo0:
Is it in the roundstart vote?

Ilya246:
> Is it in the roundstart vote?

yes, also apparently POIs for some reason manually specify a whitelist of what gamemodes they should spawn in
fixed now

ark1368 (approved):

PR #2543, opened 2025-11-02
Flashpoint Upgrades

Archylle:
## About the PR
Okay so I added a few changes. I replaced the Kargil with the 140mm longbow, removed the marauder and replaced it with a Tarynx as well as added ammo so you can replace its limited stores. I reworked the thrusters so its somewhat faster and added more gyros too so the flashpoint turns faster. I furthermore added a gunloader to the cargo bay for easy reloading and made the flashpoint cost 60k less. I also fixed the broken disposal system.

## Why / Balance
The flashpoint needed a rework and while this isn't a complete rework its enough to make people want to use it more. The flashpoint is 60k cheaper, has a stronger gun on it and also has an emp and a reloader. Its more expensive than the hammerhead which I think people will be more inclined to use but I think the touchups to the flashpoint makes it a tad bit better.

## How to test
Buy the flashpoint at exped lodge and test it

## Media
image
image
image
image

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Improved the Flashpoint


OneZerooo0:
We really need a better ship to replace the flashpoint, it’s iconic but form wise it’s relatively outdated
Longbow is also really, really unbalanced right now

PR #2552, opened 2025-11-03
Laser Drill Fix

EckoAurum:
## About the PR
The laser drill technically worked fine, I assume, but noticed when changing "requireExpedition" to false, the setting didn't work.

My best guess is that the gridQuery looked for a map to see if it has a grid component. I believe only expedition maps have one of those baked in, so it would fail the check no matter the condition set if it wasn't on an expedition planet. No issue in normal circumstances, but won't let you toggle the requirement correctly in the yaml.

I shoved a bunch of or statements to replace the sanity check. cancel if there's no map, if the map doesnt have a mapcomponent, if the grid doesnt exist, or if the grid doesnt have a grid component.

Can probably cut that down, i don't know if it's physically possible to even have maps or grids without their components, but i'm not usually the redundancy type of coder, so i threw it in just in case, feel free to cut it.

## Why / Balance
I discovered this when trying to show someone how easy and simple it was to make a new laser drill that works off expeditions for their map. I was wrong, because it was broken.

I've added a prototype for a station laser drill that can be used for mapping. It's unanchorable, works on stations, and is twice as slow.

## How to test
Load up, hook up laser drill and advanced laser drill to the dev station. Watch as only the advanced one will mine.

Go to an expedition, spawn both, watch as the regular laser drill now mines.

## Media
No visual changes

## Requirements

## Breaking changes

**Changelog**
This has no noticeable impact to players.

### DIFF
--- a/Content.Server/_Goobstation/ItemMiner/Systems/PlanetMinerSystem.cs
+++ b/Content.Server/_Goobstation/ItemMiner/Systems/PlanetMinerSystem.cs
@@ -40,7 +40,7 @@ public sealed class PlanetMinerSystem : EntitySystem
         var gridUid = xform.GridUid;
 
         // check if we're on a planet
-        if (!_gridQuery.TryComp(mapUid, out var mapGrid) || gridUid == null)
+        if (mapUid == null || !_mapQuery.HasComp(mapUid.Value) || gridUid == null || !_gridQuery.HasComp(gridUid.Value))
         {
             args.Cancelled = true;
             return;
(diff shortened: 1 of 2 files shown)


Ilya246:
intended, the component is supposed to make it only work on planets
if you want it to mine offplanet just don't have PlanetMiner

Resources/Prototypes/_Mono/Entities/Structures/Machines/laser_drill.yml
Ilya246, line 1: just make a BaseLaserDrill with components you want on both and make both drill protos share it as parent
don't have PlanetMiner on the station drill

Content.Server/_Goobstation/ItemMiner/Systems/PlanetMinerSystem.cs
Ilya246, line 1: don't need this, i said why

Ilya246 (changes requested):

EckoAurum:
> intended, the component is supposed to make it only work on planets if you want it to mine offplanet just don't have PlanetMiner

Oh, you know, i don't know why i had convinced myself the ItemMiner component was in there

Ilya246:
also it's fine to add protos for your map in the same PR as you add the map

PR #2553, opened 2025-11-03
ADS Weapons Re-Buff

Nope1356:
Yea.
People complained too much, then powercreep kept chugging while ADS weapons got nerfed.
This rebuffs ADS weapons into being something scary and valuable, not fucking analogs of 90mm and kargil.

To whoever reviews this: Do not crucify me for touching the explosions. I based my knowledge off of the other explosions in the code.

edit: unbuffed harbinger since I had people complain about that specifically. Did not touch the others since the EMP seemed to be main point of contention.

**Changelog**

:cl:
- tweak: Buffs 3/5 ADS weapons with one or more of the following: increased range, damage, and fire rate.

### DIFF
--- a/Resources/Prototypes/_Mono/Entities/SpaceArtillery/SpaceArtillery/Energy/launcher.yml
+++ b/Resources/Prototypes/_Mono/Entities/SpaceArtillery/SpaceArtillery/Energy/launcher.yml
@@ -666,7 +666,7 @@
   - type: WirelessNetworkConnection
     range: 500
   - type: Gun
-    fireRate: 0.025 # 40s
+    fireRate: 0.04 # 25s same as reaper the nebula gets a treat
     minAngle: 0
     maxAngle: 0
     projectileSpeed: 750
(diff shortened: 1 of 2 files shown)


Avalon-Proto:
I will mention the Harbringer had a slower fire rate to prevent permastuun via EMP spam

Nope1356:
> I will mention the Harbringer had a slower fire rate to prevent permastuun via EMP spam

Considering the relatively light armament and armor of the nebula, and current tactics employed against ADS involving multiple ships...
It is primarily done because a 40s reload on an EMP weapon that only works ~45% of the time isn't useful in any meaningful capacity. It can come in clutch when used against smaller ships like vega, calypso, dagger, anaconda, etc. But those ships regularly move in a group, which invalidates your single-target EMP

OneZerooo0:
Nope! ship guns in general just need a comprehensive rebalancing, not an arms race for what gets buffed

PR #2536, opened 2025-11-01
New Scrapyard Ship: Mite

Ilya246:
## About the PR
adds new ship to scrapyard console
has MIL IFF
costs 45K because it's pretty much a pile of scrap and it isn't good
is intentionally questionable
no company designation because see description i wrote for it 
also reduces PTK cost 1500->400 because why was it 1500

roster: small salvage brawler

## Why / Balance
good
and i think we need some ships like this because they're fun to have

## Media
image
image

**Changelog**

:cl:
- add: Added Mite, a new military ship, to scrapyard consoles.

PR #2557, opened 2025-11-04
Ruby Price Fix

Ilya246:
## About the PR
120k->125k

## Why / Balance
if i appraisegrid an init'd ruby ingame it appraises to 123k, and price being low also for some reason fails test on #2536 specifically and nowhere else
i have no idea how this happens and why other PR tests are succeeding

**Changelog**

:cl:
- tweak: Ruby price 120k->125k to raise price above worth.


bluebrouny:
damn must have been a change cuz it gave under 100k when i appraised in game during the fix

Avalon-Proto:
> damn must have been a change cuz it gave under 100k when i appraised in game during the fix

Probably forgot to mapinit first. Happens

PR #2575, opened 2025-11-09
Drillsite Fixes

Ilya246:
## About the PR
- can't unanchor the drills anymore
- now has ghost warp point

**Changelog**

:cl:
- fix: Derelict Drillsite drills can no longer be unanchored, and it now has a ghost warp point.


Ilya246:
easily reviewable fix so i'll just

DoubleRiceEddiedd:
Seems like people still able to unanchor it by removing the floor tile the drill was on.

PR #2576, opened 2025-11-09
Fix Master Testfail

Ilya246:
## About the PR
fails MaterialComposition price check and for some reason on my PR (#2514) specifically and no other


Ilya246:
holy heisentest

Ilya246:
okay i think this is the same issue that's causing `invalid`s in links in our map files and i have no idea how to fix it or what's even causing it
time to rerun test

PR #2588, opened 2025-11-10
Hyperconvection't

Ilya246:
## About the PR
see changelog
lathe worker rights no more sweatshops

## Why / Balance
too strong
can also put it in space and negate its main downside
not anymore
matter bins are too strong too so nerfs them too

## How to test
see media

## Media
image

**Changelog**

:cl:
- tweak: Hyperconvection upgrade material usage 0.5x -> 0.75x.
- tweak: Hyperconvection lathes will now refuse to work if it too hot (>100C) or can't deposit enough heat (if it's in space).
- tweak: Matter bin effect on lathes nerfed.

### DIFF
--- a/Content.Server/Lathe/Components/LatheHeatProducingComponent.cs
+++ b/Content.Server/Lathe/Components/LatheHeatProducingComponent.cs
@@ -25,9 +25,10 @@ public sealed partial class LatheHeatProducingComponent : Component
     // Mono
     /// <summary>
     /// Refuse to work if depositing the energy would bring the air to above this temperature.
+    /// Doesn't apply if null.
     /// </summary>
     [DataField, ViewVariables(VVAccess.ReadWrite)]
-    public float TemperatureCap = Atmospherics.T0C + 100f;
+    public float? TemperatureCap = null;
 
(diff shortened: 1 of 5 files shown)


Ilya246:
hold up needs fixes

Ilya246:
fixes real

Danchi299:
Limiting at which temperatures the upgrades can work?
Understandable so that the sidegrades can work their purpose (and probably kill less people along the way).
Nerfing the said sidegrades AND matter bins?
Nope, you lost me.

Ilya246:
> Limiting at which temperatures the upgrades can work? Understandable so that the sidegrades can work their purpose (and probably kill less people along the way). Nerfing the said sidegrades AND matter bins? Nope, you lost me.

there's currently a problem where we can't balance the economy properly because either:
a) we balance for unupgraded lathes, in which case upgraded lathes can make infinity things
b) we balance for upgraded lathes, in which case unupgraded lathes can't make anything
i've opted for just reducing the gap so we can actually balance

it's just the similar atmos problem all over again

OneZerooo0 (approved):

PR #2590, opened 2025-11-10
Kill Circle Shader

Ilya246:
## About the PR
don't need it
also makes all radars pannable

## Why / Balance
good

## Media
image

**Changelog**

:cl:
- tweak: Radars are now square.
- tweak: All radars are now pannable.


ark1368 (changes requested):
image

Arrows still act like the circle shader is there. Unless you intended that

RangerXVII:
> Arrows still act like the circle shader is there. Unless you intended that

I think that's what it's like on frontier.

Ilya246:
> Arrows still act like the circle shader is there. Unless you intended that

i think it's better this way since starsector-style arrows on the rectangular border are hard to read

ark1368 (approved):

PR #2594, opened 2025-11-10
Pilotable Derelicts

Ilya246:
## About the PR
- kills StationDampenComponent
- makes it possible to switch dampening on derelict ADS zenith POI
- makes it possible to drive event derelicts

## Why / Balance
good
some of the derelicts are also pretty much just ships

## How to test
tp to zenith POI
see you can switch its dampening

trigger bluespace salvage event
repair derelict
you can actually drive it

## Media
image

**Changelog**

:cl:
- tweak: You can now repair and drive bluespace salvage event derelicts.
- fix: You can now set dampening mode on the derelict ADS Zenith.


OneZerooo0 (approved):
seems fine, not sure if they despawn or not though

PR #2612, opened 2025-11-12
1984

Ilya246:
no more cutting tile under it to steal

**Changelog**

:cl:
- fix: Advanced laser drills can no longer be re-anchored once loose.


Ilya246:
being stolen every round

PR #2616, opened 2025-11-12
Hackable CC Vends

Ilya246:
## About the PR
you can now hack vendors in safezones

## Why / Balance
why is it not allowed
contraband inventory is all just fun items or items with no real gameplay effect or which you can obtain easily

## Media
image

**Changelog**

:cl:
- tweak: Safezone vendomats are now hackable.


ark1368 (approved):

Danchi299:
> why is it not allowed

Most concerning reasons are usually:
- Breaking into security for items
- Powering down all vendomats making them unusable
- Getting free stuff (limited to 3 items max from Frontier changes)

Probably not a big deal.

Ilya246:
PR only affects vends not doors, and the rest are a nonissue yeah

PR #2642, opened 2025-11-16
Giga Cleanup Overhaul

Ilya246:
## About the PR
- kills other cleanup system i didn't notice which should've been superceded by new grid cleanup
- (probably?) makes nearby players check run faster
- now checks for anything to what a player can return nearby, so no deleting anything near revivable dead people or brains
- refactors every cleanup system
- makes cleanup system lag be basically unnoticeable by deferring all the checks over a long duration
- replaces cheap entity cleanup system with a general spaced entity cleanup system
- also now no longer deletes anything close to a grid except other grids

## How to test
- test that grid cleanup works
- test that cheap entities in space get deleted if there's no players or bodies/brains with bound ghosts nearby

**Changelog**

:cl:
- fix: Overhauled entity cleanup systems. Report weirdness with things disappearing when you go away for a bit.


OneZerooo0 (approved):
Looks good but I’m not too familiar with cleanup code

Ilya246:
> Looks good but I’m not too familiar with cleanup code

at this point the whole thing is almost entirely mine so of course you aren't
i tested it pretty well though and even benchmarked it, doesn't lag

PR #2669, opened 2025-11-22
Cryochem RR Reversal Redux

tonotom1:
## About the PR

Adds HealthScaleEffect.cs and implements it into Aloxadone, Stelloxadone, and Traumoxadone as a more elegant means to achieve RR removal. Enjoy being able to undo tens of thousands of damage over minutes instead of hours.

I am not proud of this. Do not credit me.

## Why / Balance

This is dedicated to the radiation crate-RRed mailman who was based enough to wait TWO HOURS for a revive by being healed with nothing but cryoxadone because adv cryo chems had the worst effort to effect ratio imaginable. Now they might actually be worth batting an eye for.

## How to test

There are commented-out loggers left in the code. Feel free to uncomment them and test the cryochems.

Hit yourself with a bang stick burner/gibber or do something like damage Toxin 100000 true to yourself for convenient damage.

Aghost, drag your corpse into space, and then use solution manager on it.

## Media

image

## Requirements

## Breaking changes

I dared to add something congruent to the health system, which is a garish rat's nest. REVIEW WITH GREAT CAUTION.

**Changelog**

:cl:
- tweak: Aloxadone, Stelloxadone, and Traumoxadone round removal reversal effect is much more potent now, scaling based on the current damage of the metabolizer.


Content.Server/_Mono/EntityEffects/Effects/HealthScaleEffect.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// nonzero chance someone might suddenly decide to port a different behavior over.
+        /// The ENTIRETY of the health system is a rat's nest and the less technical debt I risk for this effect the better
+        /// TRUST ME: the yaml for specifying this properly is ugly but it's better than having to fix this
+        /// </summary>
+        [DataField(required: true)]
+        [JsonPropertyName("healthscale")]
+        public DamageSpecifier HealthScale;
+
+        [DataField]
+        [JsonPropertyName("ignoreResistances")]
```
Ilya246: why JsonPropertyName
just [DataField] and done

Content.Server/_Mono/EntityEffects/Effects/HealthScaleEffect.cs
```diff
@@ -0,0 +1,133 @@
...
+            var damageSystem = args.EntityManager.System<DamageableSystem>();
+            var damageable = args.EntityManager.GetComponentOrNull<DamageableComponent>(args.TargetEntity);
+
+            if (damageable == null)
+                return;
+
+            var currentDamage = damageable.Damage;
+
+            //DEBUG: Log args of interest
+            //Logger.Info($"ReagentQuantityArg: {(reagentArgs != null ? reagentArgs.Quantity.ToString() : "N/A")}, ReagentScaleArg: {(reagentArgs != null ? reagentArgs.Scale.ToString() : "N/A")}");
```
Ilya246: just remove those

Content.Server/_Mono/EntityEffects/Effects/HealthScaleEffect.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            foreach (var (type, currentAmount) in currentDamage.DamageDict)
+            {
+                if (HealthScale.DamageDict.TryGetValue(type, out var scaleCoeff))
+                {
+                    var applyscale = reagentArgs != null ? reagentArgs.Scale : 1.0f;
+                    var adjusted = currentAmount * (scaleCoeff - 1.0f) * applyscale;
```
Ilya246: shouldn't this be `currentAmount * MathF.Pow(scaleCoeff - 1.0f, applyscale)` so it properly applies scaling

Resources/Prototypes/_NF/Reagents/medicine.yml
```diff
@@ -27,15 +28,17 @@
 #            Slash: -2
           groups:
             Brute: -18
-      - !type:HealthChange
+      - !type:HealthScaleEffect
         conditions:
         - !type:Temperature
           max: 213.0
         - !type:TotalDamage
           min: 1000
```
Ilya246: you can set this to 200
same below

Ilya246 (changes requested):

tonotom1:
Cool review, glad it's not turbo fucked beyond saving. I'll get around to addressing your findings after some sleep and a grocery run

Content.Server/_Mono/EntityEffects/Effects/HealthScaleEffect.cs
tonotom1: reagentArgs.Scale is a value from 0 to 1 that corresponds to chem quantity metabolized vs the full metabolism rate of the chemical. IE If a default 0.5u/t chem only had 0.25u left to metabolize, reagentArgs.Scale would be 0.5

I multiply the health adjistment amount by reagentArgs.Scale to linearly reduce the effect of the chem if a partial dose is injected, to eliminate potential homeopathy shenanigans.

Using the value as an exponent would have wacky effects where homeopathic healing is conceptually real instead of eliminated

Ilya246: > reagentArgs.Scale is a value from 0 to 1 that corresponds to chem quantity metabolized vs the full metabolism rate of the chemical. IE If a default 0.5u/t chem only had 0.25u left to metabolize, reagentArgs.Scale would be 0.5
> 
> I multiply the health adjistment amount by reagentArgs.Scale to linearly reduce the effect of the chem if a partial dose is injected, to eliminate potential homeopathy shenanigans.
> 
> Using the value as an exponent would have wacky effects where homeopathic healing is conceptually real instead of eliminated

no, that's wrong, exponent is still proper, since, say, if we take `0.8` as the scaleCoeff and `h` as damage, we get:
1x scale: `h = h * 0.8` after 1 tick, `h = h * 0.64` after 2 ticks
0.5x scale, linear: `h = h * 0.9` 1t, `h = h * 0.81` 2t
0.5x scale, exponent: `h = h * sqrt(0.8)` 1t, `h = h * 0.8` 2t
with exponent, 2 applications of 0.5x scale is the same as 1 application of 1 scale, which it should be

tonotom1: I tried it anyway to satiate my 'tism and MathF.Pow threw a fit over the arguments involved (negative base, fractional power)

Ilya246: > I tried it anyway to satiate my 'tism and MathF.Pow threw a fit over the arguments involved (negative base, fractional power)

what'd you do exactly?
should be `var adjusted = currentAmount * MathF.Pow(scaleCoeff - 1.0f, applyscale);`

Ilya246: actually yeah no that'd lead to negative base and it's just wrong, try
`var adjusted = currentAmount * (MathF.Pow(scaleCoeff, applyscale) - 1.0f);`
scalecoeff is the `0.8` in the example above so we have to pow it

tonotom1: Oh I get the intent now that you've supplied that formula and supplied the example

Pardon the .float() slapped in because fixedpoint moment

Ilya246 (approved):

PR #2732, opened 2025-12-03
The Axe Comes Down - Removal of Random Chimera Spawns

Avalon-Proto:
## About the PR
This PR totally removes the ability for chimera to spawn randomly, their rules to do so, and 4 of the shuttles that needed removal as per starch's request.

## Why / Balance
Chimera are becoming TOO common a sight in the sector. While some threat is nice, it's not fun when literally every round has dozens of chimera running around, seal clubbing new players, and killing whoever they please. Chimera should be a very are and limited thing to see, only happening when someone intentionally uses the bioweapon.

## How to test
Load up, see its gone

## Requirements

## Breaking changes

**Changelog**

🆑 
- remove: Game Votes have been disabled
- remove: Chimera no longer randomly spawn.
- remove: Chimera shuttles removed as per starch's request


HvyEarly:
My only complaint is that use of the bioweapon is quite common in and of itself. While this will certainly reduce the amount of chimera sightings, it may not be enough to make them truly rare (stated goal of the rework)

Avalon-Proto:
> My only complaint is that use of the bioweapon is quite common in and of itself. While this will certainly reduce the amount of chimera sightings, it may not be enough to make them truly rare (stated goal of the rework)

IT will at least make them more controlled, and give people the ability to stop the outbreaks from happening instead of a new source popping up every 30 minutes.

Avalon-Proto:
@Ilya246 I have no idea what has caused these test fails when I thought I removed all mentions of both game modes, and opening those files shows the same

Ilya246:
> @Ilya246 I have no idea what has caused these test fails when I thought I removed all mentions of both game modes, and opening those files shows the same

probably mentioned in the base poi proto

OneZerooo0:
I like random chimera spawns but they do seem to be too frequent

OneZerooo0:
Closing this because we need an *actual* solution, not just removing all chimera spawns - you can reopen it if you want to just make all rounds mixed

PR #2749, opened 2025-12-06
Kill Spdx

Ilya246:
## About the PR
kills every spdx header

## Why / Balance
spdx is evil and never should have been used
direct request from durk to do this


NazrinNya:
what

PR #2754, opened 2025-12-06
Ship Movement Rework

Ilya246:
## About the PR
refactors ship movement
changes the way max velocity limiting works
max velocity is now limited on an ellipse
if you're going at or above max velocity, going to the side now properly just rotates your velocity instead of not letting you move, so if you go forward at max speed and thrust to the side it will let you

## Why / Balance
necessary for ship HTN

## How to test
1. get pluto
2. go forward at max speed
3. go right
4. observe it actually go right

## Media
[x] tested, works

**Changelog**

:cl:
- fix: Ships going at max speed can now properly thrust to the side without having to brake first.
- tweak: Decreased thruster upgrade effect on max speed. There's now only effect via thrust increase.

### DIFF
--- a/Content.Server/Physics/Components/PilotedShuttleComponent.cs
+++ b/Content.Server/Physics/Components/PilotedShuttleComponent.cs
@@ -1,0 +1,18 @@
+using System.Numerics;
+
+// Mono - whole file
+
+namespace Content.Server.Physics.Controllers;
+
+/// <summary>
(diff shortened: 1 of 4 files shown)


Ilya246:
ok don't merge yet i'll tweak some more

Ilya246:
done, tweaked more

OneZerooo0 (approved):
looks good but needs second review

OneZerooo0:
fuck it we ball

PR #2783, opened 2025-12-10
Update StrippableBoundUserInterface.cs

core-mene:
resize the stipping UI from 220 to 300 so you can see every icons

(ignore icons not being the monolith ones)
image

:cl:
- fix: Fixed part of stripping UI being hiden.


NazrinNya (approved):
there were more icons fr?

core-mene:
> there were more icons fr?

you see the issue now :clueless:

PR #2790, opened 2025-12-11
Researchable Advanced Tactical Bags

tonotom1:
## About the PR

Adds advanced tactical bags behind a 350k point universal faction research with T3 materials prerequisite.

## Why / Balance

Player requested

It's this expensive to avoid infringing upon NT crate loot too hard, but still allow someone jonesing for an inventory capacity upgrade to work towards it without RNG.

## How to test

Research and check mercfab + milfab

## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Advanced tactical bags can now be researched at a high cost.

### DIFF
--- a/Resources/Locale/en-US/_Mono/research/faction-universal.ftl
+++ b/Resources/Locale/en-US/_Mono/research/faction-universal.ftl
@@ -11,3 +11,5 @@ research-technology-universal-advanced-light-fighter-ordinance = Advanced Fighte
 
 research-technology-universal-armorplates-1 = Ballistic Composites Reverse Engineering
 research-technology-universal-armorplates-2 = Plasteel Hex-ply Stamping Reverse Engineering
+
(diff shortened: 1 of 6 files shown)


NazrinNya (changes requested):
In the current state of research, this PR is very controversial. 
Having bunch of research that's locked under grindwall isn't a good practice. One or two times? Sure, but having extremely expensive research just because one minmaxxer wanted that shouldn't be a thing. Those bags aren't even that good to cost 500k points.
Either move it to the less-rare tech disk or just dont add it to research at all.

tonotom1:
I do not believe that there's any means to carry forward with this change in a satisfactory manner. If anybody comes up with a solution that doesn't infringe on NT crate loot too hard, or redundantly taps into gamerloot (since NT crates already exist), don't hesitate to comment.

PR #2802, opened 2025-12-15
Hide+farify Faction Bases

Ilya246:
## About the PR
hides halc and helios
moves helios to 12-13k, halc 14.5-16k

## Why / Balance
see all the discord discussions

**Changelog**

:cl:
- tweak: Both faction bases are now hidden and spawn far out.

### DIFF
--- a/Resources/Prototypes/_Mono/PointsOfInterest/pdvhelios.yml
+++ b/Resources/Prototypes/_Mono/PointsOfInterest/pdvhelios.yml
@@ -9,14 +9,15 @@
   id: HeliosFortress
   parent: BasePOI
   name: 'PDV Helios Fortress'
-  minimumDistance: 7500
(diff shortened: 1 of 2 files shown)


NazrinNya (approved):
https://github.com/user-attachments/assets/68d9900c-db8d-4170-8a04-1465c35c4b57

OneZerooo0 (approved):
may want to reduce base armament

jmlop0:
Why are we doing this when you also have a poll ongoing about off sector bases? Where are the discussions you’re referring to so I can read them if possible?

Ilya246:
> Why are we doing this when you also have a poll ongoing about off sector bases? Where are the discussions you’re referring to so I can read them if possible?

the poll exists more so for the discussion and for what we should do in the future
discussion is in poll thread

PR #2804, opened 2025-12-15
Contraband Rework

Ilya246:
## About the PR
- kills all contraband
- adds new generic TSF/PDV/other/other+taxed categories for faction items
- puts a few faction items under contra status

this will need further tweaks and marking of way more items, but i'd say we merge this as-is so people/other maints/etc can PR it themselves
currently this only marks hardsuits, a few weapons, and contra crates

## Why / Balance
contra doesn't make sense
and we need both factions to be able to hand in other faction gear

## Media
image

**Changelog**

:cl:
- tweak: Reworked contraband. Almost all items are no longer contraband, and instead some items are marked as faction gear and can be handed in by the opposite faction. More will be marked soon.


OneZerooo0 (approved):

OneZerooo0:
Speedmerge since faction related - we need to improve them

PR #2852, opened 2025-12-23
Construction Bag/Box Tag Additions

Helicaste:
## About the PR

Adds the Recyclable tag to the item whitelist for Construction Box and Construction Bag, which also applies to the BS variant.

## Why / Balance

This allows construction boxes to contain items with the Recyclable tag. This does not impact the balance of recycling those items, as their material value is unchanged. It does effect potential storage density for some items.

This only expands the list to items that are explicitly recyclable. It apparently does not apply to items that gain the tag from a parent, particularly a multi-parent like Punk Goggles which should in theory fit in the box but actually don't. In testing this also applies to ammo boxes/magazines, which are recyclable but also don't fit. The only other explicitly recyclable item is AME fuel jars.

The most materially dense pieces of scrap the box can now transport are already 4x4 allowing the construction box to store at most 10 of them. That's at most 300 steel or less of any other material per collection trip.

The same applies to the construction bag and bluespace variation except for their inability to store Large and Huge items.

I do not think a special box explicitly for storing scrap is required here, that'd just be mapping bloat and mildly irritating to interact with. This does decrease the effectiveness of swinging a construction box around to collect materials from debris crates by virtue of increasing the potential number of things picked up. This does increase the sidegrade value of using the bags because they exclude the largest chunks of scrap.

#Draft
-intend to add flatpack tag to BaseFlatpack to allow construction bag/box to store flatpacks.
-might buff BS CB to include Large items.
-maybe blacklist AME fuel jars for explosion reasons.

## How to test

Spawn construction bag/box, spawn recyclable scrap items nearby, it will be pulled in by the magnet.
## Media

N/A
## Requirements

## Breaking changes

N/A

**Changelog**

:cl:

- tweak: Construction Bags and Boxes can now store scrap items found on debris asteroids/vgroids.

### DIFF
--- a/Resources/Prototypes/_NF/Entities/Objects/Specific/Engineering/construction_bag.yml
+++ b/Resources/Prototypes/_NF/Entities/Objects/Specific/Engineering/construction_bag.yml
@@ -39,6 +39,8 @@
       - RodMetal1
       - Sheet
       - CableCoil
+      - Recyclable # Mono
+      #DRAFT: Add flatpacks
       components:
       - ConstructionMaterials
     blacklist:
--- a/Resources/Prototypes/_NF/Entities/Structures/Storage/construction_box.yml
+++ b/Resources/Prototypes/_NF/Entities/Structures/Storage/construction_box.yml
@@ -41,6 +41,8 @@
       - RodMetal1
       - Sheet
       - CableCoil
+      - Recyclable # Mono
+      #DRAFT add flatpacks
       components:
       - ConstructionMaterials
       - ResearchDisk # Mono


DoomAngelBlade:
Uhhhh.
Construction box/bag literally is for... Construction mats.
Steel, plasteel, plasitanium, uranium, glass, rods, wires.

Now it's a junk box/bag?

Can we just have a separate junk box/bag?

Helicaste:
> Uhhhh. Construction box/bag literally is for... Construction mats. Steel, plasteel, plasitanium, uranium, glass, rods, wires.
> 
> Now it's a junk box/bag?
> 
> Can we just have a separate junk box/bag?

I'm not convinced about needing a new bag/box only for scrap, which is why this isn't that. It makes enough sense for scrap to fit in the construction containers that i'm not concerned about that. This doesn't remove the function for the construction container to contain construction materials, you're still entirely capable of doing that. Adding a full new box seems like bloat.

In short, adding a new set of containers only for scrap seems like more effort than is warranted. They'd only exist for managing scrap when my intention is making scrap easier to manage.

tonotom1:
When you make a tool for everything without the room to accomodate for everything and leave no means of specializing with the tool, it becomes a tool for nothing.

DoomAngelBlade:
Construction bags/boxes have one great use for salvagers - gather all mats found in wrecks, and there are mats on wrecks to be found.

Now, if this gets merged, it's either NOT going to be used or be extremely annoying to have to constantly stop, toss out all the recyclables for you to continue gathering mats on wrecks.

This is going to kill a thing and fix for it would be... to "Bloat"...

NazrinNya:
we shouldn't make tools for literally everything

Helicaste:
> When you make a tool for everything without the room to accomodate for everything and leave no means of specializing with the tool, it becomes a tool for nothing.

That's very cool but I would advise thinking about the changes in-game instead of the abstraction you've reduced them to. Functionally, a single construction box is still able to hold the entire contents of a large debris asteroid without any issue. There is some variability if a lot of Large-size scrap items spawn, but not very much.

Helicaste:
> Construction bags/boxes have one great use for salvagers - gather all mats found in wrecks, and there are mats on wrecks to be found.
> 
> Now, if this gets merged, it's either NOT going to be used or be extremely annoying to have to constantly stop, toss out all the recyclables for you to continue gathering mats on wrecks.
> 
> This is going to kill a thing and fix for it would be... to "Bloat"...

You have accurately pointed out why it would be bloat. It’d be another item to map for scrap/mining ships, another flatpack in the salvage vendor, another thing to micromanage drop-off trips to your processors, possibly another research node to unlock, and all to accommodate a feature people aren’t using. As I said, my intent is to make scrapping easier to manage. Adding another box is adding another step most people won’t take, hence: bloat.

tonotom1:
> That's very cool but I would advise thinking about the changes in-game instead of

My brevity was not absence of consideration:

1: I want to use construction bags because I don't like dragging a construction box. This change guts construction bags, which already barely fulfill the capacity requirements to handle wrecks for solely materials. Reducing the capability of solutions when there was previously more options is not more sand in the sandbox.

2: Not everybody wants to or has the equipment to reclaim scrap. Railroading people into managing items they don't want is once again, not more sand in the sandbox.

3: People use construction storage for its namesake ofconstruction as well. Overgeneralizing the magnet impedes this function and induces gameplay chaff requiring corrective action if a magnet is mistakenly left on near an an active scrap reclaimer pile. When it involves many items, the doafter to dump becomes punishing. Once again, similar vein: not good for the sand in the box.

Once again, I do not agree that adding this without a new bag to tie it to would be a good change. It would overgeneralize its capabilities and thus result in not being *exactly* what people need.

DoomAngelBlade:
> You have accurately pointed out why it would be bloat.

Exactly.
And you're making that bloat necessary.

tonotom1:
Furthermore, after rumination while having a delicious custard tart: I believe it to be possible to implement the intent of this change with an additional proto without inducing the aforementioned bloat and overgeneralization issue.

One could create a new crate prototype with magnet behaviour as-is, and an associated flatpack.

Initialized ships could hold crate flatpacks instead of the crates themselves and they can subscribe to whichever activity they wish to commit to simply by unpacking the crates they desire. This way, anyone can choose between mining, materials-only salvage, and full-variety salvage.

My proposal, even in isolation of your proposed PR has the added benefit of reducing from our current footprint of boxes occupying two spaces to a single decorative rack holding the flatpacks. Furthermore, unobstructed subscription to desired activity goals is also achieved via lathe recipes for the bags and  crates.

Helicaste:
A new scrap box would pair well with revisions to the scrap system. It also apparently exists as an upstream PR from Frontier probably coming soon to own on video and DVD.

PR #2865, opened 2025-12-25
New Rammer Drone

Ilya246:
## About the PR
add new outer reach rammer drone: medusa

has really strong forward thrust but one small gyro for rotation
non-leading AI core, but ignores obstacles so it's basically just a bullet

## Why / Balance
we need more drones beyond the 3 we have

## Media
image
thrusters are fixed in latest update
image

**Changelog**

:cl:
- add: Added a new type of rammer drone to the outer reaches.


DoomAngelBlade:
Where's the lootbox?
(or 2?)

Ilya246:
lower screenshot is a bit outdated and doesn't show it
also this thing can't turn it doesn't need 2

thoughtlessuser:
evil ass dart that will kill you if you leave

Arcticular1 (approved):
Seems all good.

PR #2886, opened 2025-12-28
Refactor Hydrakin

beck-thompson:
## About the PR

Title! It will basically act the same as the other system but is cleaned up. Mainly I split off the passive heating that they have into a new component which is a little clearer and also made their heating reduction ability clearer (Added comments and changed the equation around so its easier to read). They do seem to heat up slightly faster on the top end but its not noticeable much. I also moved some of the files around to be in the correct folder location (Components where inside systems)


Content.Server/_Obelisk/Species/Components/HydrakinComponent.cs
```diff
@@ -0,0 +1,27 @@
+using Content.Shared.Actions;
+using Robust.Shared.Audio;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server._Obelisk.Species.Components;
```
Ilya246, line 5: _Obelisk? also the directory name

Content.Server/_Obelisk/Species/Components/PassiveHeatGenerationComponent.cs
```diff
@@ -0,0 +1,51 @@
...
+namespace Content.Server._Obelisk.Species.Components;
+
+/// <summary>
+/// Passively adds or takes away heat from the entity.
+/// </summary>
+[RegisterComponent, AutoGenerateComponentPause]
+public sealed partial class PassiveHeatGenerationComponent : Component
+{
+    /// <summary>
+    /// Amount of jules to add or subtract every <see cref="UpdateInterval"/>
```
Ilya246: jules

Content.Server/_Obelisk/Species/Components/PassiveHeatGenerationComponent.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// If the heat will ignore heat resistance or if it will be taken into account.
+    /// </summary>
+    [DataField]
+    public bool IgnoreHeatResistance;
+
+    /// <summary>
+    /// <see cref="Joules"/> will be applied with a period of this time.
+    /// </summary>
+    [DataField]
+    public TimeSpan UpdateInterval = TimeSpan.FromSeconds(1);
```
Ilya246: just change the heat generation to watts, get rid of this update interval, and instead add an update interval to the whole system so it only runs the EQE every second
also just use an accumulator (+= TimeSpan.FromSeconds(frameTime)) for that, better than the NextDoThingAt pattern because it's more reliable

Ilya246 (changes requested):

Content.Server/_Obelisk/Species/Components/HydrakinComponent.cs
beck-thompson: Whats the issue? This was how it was before 🤔
Isn't it the idea to put stuff in the forks folder so its organized?

Ilya246: are they obelisk?

Ilya246 (approved):
seems fine

NazrinNya (approved):
shouldn't change anything as it only cleans up code

PR #2899, opened 2025-12-30
Radar Edge Markers

Ilya246:
## About the PR
see media, self-explanatory

## Media
image

image

**Changelog**

:cl:
- add: Added radar edge markers for mapping.


Skyguard12:
Doesn't Mass Scanner already do this by itself?

Ilya246:
> Doesn't Mass Scanner already do this by itself?

you put those inside grids

Skyguard12:
> > Doesn't Mass Scanner already do this by itself?
> 
> you put those inside grids

I don't think I understand what you mean. Do you want some random lines floating in space as opposed to being on grids?

Ilya246:
> I don't think I understand what you mean. Do you want some random lines floating in space as opposed to being on grids?

you get lines inside grids and not on their edges

NazrinNya (approved):
goida

PR #2910, opened 2025-12-31
Fix Burst Shipguns

Ilya246:
## About the PR
they currently fire one burst and stop

**Changelog**

:cl:
- fix: Fixed Cyrexa and AK570 only firing one burst then ceasing to work.


Ilya246:
selfmerging since important

PR #2919, opened 2026-01-01
Various Projectile Fixes

Ilya246:
## About the PR
attempts to fix:
- projectile phasing
- reports of projectile velocity inheritance being bad
- projectiles having linear dampening

2 of those are likely heisenbugs with unknown cause but i tried to squash any possible cause, and couldn't replicate them any more in localhost however i tried

## Media
subnuke charon hit
image

max speed moving europa firing
image

**Changelog**

:cl:
- fix: Projectiles now properly no longer have velocity falloff.
- fix: Projectiles should no longer phase through armor.
- fix: Potentially fixed reported weirdness with projectiles inheriting grid velocity.


NazrinNya (approved):
should be fine

PR #2920, opened 2026-01-01
Fix Gunnery Server on Caput

NazrinNya:
## About the PR

Caput had to use medium-power gunnery server, not low-power gunnery server.


NazrinNya:
self merging since its very small fix

PR #2954, opened 2026-01-05
New T2-1 Drone

Ilya246:
## About the PR
adds Gust, new outer drone
has a smart non-orbiting core that dodges your shots
also does C# tweaks to let it rotate properly

## Why / Balance
need more drones still

## Media
image
image

**Changelog**
i don't think those even need CLs because people will read this and forget


Arcticular1 (approved):
Balancing looks good

PR #2975, opened 2026-01-07
Fix Radar Markers Offset

Ilya246:
## About the PR
no longer offset by half a tile

## Media
image

**Changelog**

:cl:
- fix: Radar edge markers are no longer offset by half a tile.


Ilya246:
selfmerging since no reason not to

PR #2984, opened 2026-01-08
Delete Offgrid Markers

Ilya246:
## About the PR
all markers are now deleted when offgridded

## Why / Balance
why are they not currently being deleted
this adds this to basemarker but i don't think it's going to break anything, in fact it's probably going to fix potential issues

## How to test
1. ram baeg into grid
2. no floating vacuum markers

**Changelog**

not player-facing


NazrinNya (approved):
should be fine

PR #3072, opened 2026-01-19
Gun Executions

kotobdev:
## About the PR

Ports gun executions from DV, allowing you to execute other people *and* yourself with a variety of firearms. Emphasis on the "yourself" part if you're clumsy.

## Why / Balance
I wanted to do a gimmick as a TSF pilot and was surprised to find out you can't stick a gun in your mouth already. So, this fixes that. Only notable changes from DV's system is that executing yourself is back, and doing so is substantially faster then executing somebody else.

evil ancient non-predicted code but it works. at this point it's possible that *maybe* all dependent systems have been made Shared, but no idea if that's also true here and I'd have to rewrite the entire thing and that takes time

## How to test
grab gun (revolver, pistol, rifle)
right click urist -> no execute
right click cuffed/crit urist -> execute
right click self -> always execute

shotguns don't deal a whole lot of damage on execute since it only accounts for one pellet, but it's whatever

## Media
image
image
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: You can now execute people, including yourself, with firearms!

### DIFF
--- a/Content.Server/_DV/Execution/ExecutionSystem.cs
+++ b/Content.Server/_DV/Execution/ExecutionSystem.cs
@@ -27,7 +27,7 @@ using Robust.Shared.Player;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Random;
 
-namespace Content.Server.Execution;
+namespace Content.Server._DV.Execution; // Mono - fix namespace
 
 /// <summary>
 /// Verb for violently murdering cuffed creatures using guns.
--- a/Content.Server/_DV/Execution/ExecutionSystem.cs
+++ b/Content.Server/_DV/Execution/ExecutionSystem.cs
@@ -1,0 +1,308 @@
+using Content.Server.Interaction;
+using Content.Server.Kitchen.Components;
+using Content.Server.Weapons.Ranged.Systems;
+using Content.Shared._Shitmed.Targeting;
+using Content.Shared.ActionBlocker;
+using Content.Shared.Clumsy;
+using Content.Shared.Damage;
+using Content.Shared.Damage.Components;
+using Content.Shared.Damage.Systems;
+using Content.Shared.Database;
+using Content.Shared.DoAfter;
+using Content.Shared.Execution;
+using Content.Shared.Interaction.Components;
+using Content.Shared.Mobs.Components;
+using Content.Shared.Mobs.Systems;
+using Content.Shared.Popups;
+using Content.Shared.Projectiles;
+using Content.Shared.Verbs;
+using Content.Shared.Weapons.Hitscan.Components;
+using Content.Shared.Weapons.Ranged;
+using Content.Shared.Weapons.Ranged.Components;
+using Content.Shared.Weapons.Ranged.Events;
+using Content.Shared.Weapons.Ranged.Systems;
+using Robust.Shared.Audio;
+using Robust.Shared.Audio.Systems;
+using Robust.Shared.Player;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Random;
+
+namespace Content.Server.Execution;
+
+/// <summary>
+/// Verb for violently murdering cuffed creatures using guns.
+/// </summary>
+/// <remarks>
+/// Upstream won't have this for years so it goes in the DeltaV folder.
+/// </remarks>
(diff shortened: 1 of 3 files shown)


Content.Server/_DV/Execution/ExecutionSystem.cs
```diff
@@ -0,0 +1,309 @@
...
+        // All checks passed
+        return true;
+    }
+
+    private bool CanExecuteWithGun(EntityUid weapon, EntityUid victim, EntityUid user)
+    {
+        if (!CanExecuteWithAny(weapon, victim, user)) return false;
+
+        // We must be able to actually fire the gun
+        if (!TryComp<GunComponent>(weapon, out var gun) && _gunSystem.CanShoot(gun!))
```
Ilya246: shouldn't this be `if (!TryComp... || !CanShoot())`?

Avalon-Proto:
Just saying. You should make the damage on melee executions stupid too :blunt:

Content.Server/_DV/Execution/ExecutionSystem.cs
```diff
@@ -0,0 +1,309 @@
...
+
+        if (!CanExecuteWithAny(args.Used.Value, args.Target.Value, uid))
+            return false;
+
+        // All checks passed
+        return true;
+    }*/
+
+    // TODO: This repeats a lot of the code of the serverside GunSystem, make it not do that
+    private void OnDoafterGun(EntityUid uid, GunComponent component, DoAfterEvent args)
```
Ilya246, line 175: pretty sure this will break with exploding etc projectiles, might want to make it instead just properly use AttemptShoot or a new overload that does not give the shooter immunity to the bullets

Content.Server/_DV/Execution/ExecutionSystem.cs
```diff
@@ -0,0 +1,309 @@
...
+    [Dependency] private readonly IPrototypeManager _prototypeManager = default!;
+    [Dependency] private readonly IRobustRandom _random = default!;
+    [Dependency] private readonly IComponentFactory _componentFactory = default!;
+    [Dependency] private readonly SharedAppearanceSystem _appearanceSystem = default!;
+    [Dependency] private readonly SharedAudioSystem _audioSystem = default!;
+    [Dependency] private readonly GunSystem _gunSystem = default!;
+
+    private const float GunExecutionTime = 6.0f;
+    private const float GunSuicideTime = 2.0f; // Mono
+    private const float DamageModifier = 9.0f;
```
Ilya246: move all 3 to datafields on GunComponent

Ilya246 (changes requested):

Content.Server/_DV/Execution/ExecutionSystem.cs
kotobdev: Poked around at it and this would require *substantial* modifications to SharedGunSystem and/or general shitcode, especially when it comes to increasing projectile damage and targeting the head

Explosives do more or less work with this system, in the sense that they'll deal a bit of blunt impact damage, not go off, and become unusable. Which, when you think about it, is both kind of accurate and very funny

https://github.com/user-attachments/assets/99b5a899-8f2a-44cb-aa81-9f0208eac3f9

All other guns I bothered to test worked fine.

Ilya246: okay then

Content.Server/_DV/Execution/ExecutionSystem.cs
```diff
@@ -0,0 +1,305 @@
...
+        if (!TryComp<DamageableComponent>(victim, out var damage))
+            return false;
+
+        // You can't execute something that cannot die
+        if (!TryComp<MobStateComponent>(victim, out var mobState))
+            return false;
+
+        // You're not allowed to execute dead people (no fun allowed)
+        if (_mobStateSystem.IsDead(victim, mobState))
+            return false;
```
Ilya246: allow it

Floofbutt:
If you could make the clumsy trait increase the "shoots yourself in the foot" interaction chance that would be amazing.

Ilya246 (approved):

PR #3090, opened 2026-01-21
Further Nerf Uranium

Ilya246:
## About the PR
- allows recipes to just directly produce material so we can have fractional-production recipes
- removes gambling from uranium
- nerfs uranium production to non-insane levels
- makes reprocessing not broken (because it's still broken)

**Changelog**

:cl:
- tweak: Nerfed uranium reprocessing.
- tweak: Uranium processing is no longer RNG and now instead makes fractional amounts of uranium per recipe completion.


OneZerooo0 (approved):
RIP gambling

NazrinNya (approved):
balanced enough

PR #3118, opened 2026-01-24
Maprenderer Grid Path Support

Ilya246:
## About the PR
- can now render grids
with for example `dotnet run --project Content.MapRenderer Resources/Maps/_Mono/ShuttleEvent/wyvern.yml`

- misc fixes to rendering to make entities with sprite offset (like large thrusters) and tiles render properly

## Media
wyvern-0

**Changelog**

dev-only, none


Ilya246:
zero danger to server so selfmerging

PR #3131, opened 2026-01-27
Material Preservation

Ilya246:
## About the PR
- materials (that aren't glass) are no longer destructible because why were they
- lathes/silos now drop materials on destruction

**Changelog**

:cl:
- tweak: Lathes now drop material on destruction.
- tweak: Materials (except glass) are no longer destructible.

### DIFF
--- a/Content.Server/Materials/MaterialStorageSystem.cs
+++ b/Content.Server/Materials/MaterialStorageSystem.cs
@@ -15,6 +15,9 @@ using Robust.Shared.Audio.Systems;
 using Robust.Shared.Map;
 using Robust.Shared.Prototypes;
 
+// Mono
+using Content.Shared.Destructible;
+
 namespace Content.Server.Materials;
 
 /// <summary>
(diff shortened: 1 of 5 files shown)


Zethine07:
Crates destroy their contents held inside when broken, maybe container objects can preserve whats lost inside them on destruction too?

Ilya246:
> Crates destroy their contents held inside when broken, maybe container objects can preserve whats lost inside them on destruction too?

i don't think they do that?

Zethine07:
Last time I did tests with them and someone else, breaking the crates ended up breaking the materials inside too with explosives and such. Either its just explosions breaking the stuff inside or crates actually do delete their contents on break

Ilya246:
> Last time I did tests with them and someone else, breaking the crates ended up breaking the materials inside too with explosives and such. Either its just explosions breaking the stuff inside or crates actually do delete their contents on break

i'd suspect that yes the explosions did break the stuff inside because crates do drop stuff on destruction

Zethine07:
Then this PR should fix that issue :p ill stop here now

OneZerooo0 (approved):

PR #3171, opened 2026-02-02
Remove Self Recharge From TSF And Normal Energy Shield

mikusssssss:
## About the PR
removes the passive recharge comp from energy shields, which also counts TSF ones

## Why / Balance
If you are using an energy shield, you should not be able to consistently switch back between energy shields over and over again in a bag because they self recharge. You can hold two in your pocket. EMP'ing someone whos got an energy shield will only do something for 2 minutes, until their shield instantly recharges and can be used again. Besides this, being able to self recharge as a thing so cheap, providing 50% protection, is not something it should do. Carry a portable recharger to actually recharge your shields. Raising the price of the eshield wouldn't do anything and lowering its protection values would make it entirely useless.

## Requirements

**Changelog**
:cl:
- tweak: The base energy shield (TSF included) no longer self recharge.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Shields/shields.yml
+++ b/Resources/Prototypes/Entities/Objects/Shields/shields.yml
@@ -454,9 +454,6 @@
     - type: Battery
       maxCharge: 60
       startingCharge: 60
-    - type: BatterySelfRecharger
-      autoRecharge: true
-      autoRechargeRate: 5
     - type: RechargeableBlocking
       # WD EDIT END
 


NazrinNya:
eshield takes your hand slot, instantly invalidating any primary firearms (firing annie from 1 hand wont do jackshit).
this pr has no reasoning behind debuffing shields.

BluCottonCandy:
This entirely invalidates shields, and as already stated, having a shield already means you can't use a ton of other guns types without a huge accuracy de buff.

As for the argument of people putting them into their pockets, firstly, two shields doesn't stack.

You get the passive from one shield only, and having it in your pocket is an exploit, and bug abuse.

That isn't something we should make a PR to make worse, but instead firstly, attempt to fix the bug, and secondly, report people who are doing this.

This PR feels unnecessary, and just further makes Eshields even more clunky and difficult to use.

This PR feels like it was made soley because someone used an Eshield, and you lost the fight.

Corrupt142:
> This entirely invalidates shields, and as already stated, having a shield already means you can't use a ton of other guns types without a huge accuracy de buff.
> 
> As for the argument of people putting them into their pockets, firstly, two shields doesn't stack.
> 
> You get the passive from one shield only, and having it in your pocket is an exploit, and bug abuse.
> 
> That isn't something we should make a PR to make worse, but instead firstly, attempt to fix the bug, and secondly, report people who are doing this.
> 
> This PR feels unnecessary, and just further makes Eshields even more clunky and difficult to use.
> 
> This PR feels like it was made soley because someone used an Eshield, and you lost the fight.

if you mean the activated E shield then yeah, thats bug abuse and should be fixed although when inactive its intended to fit in a pocket slot

OneZerooo0:
womp womp

PR #3253, opened 2026-02-19
Tweaks and Balancing to Big Mecha and Power Cages/Cells

Kiithnaras:
## About the PR

- Slightly increases Sprint speed for S2 and S4 mecha to something more manageable for their large, girthy size. Still slower than a Marauder or Durand.
- Slightly increases Sabre weightless speed to something not entirely useless while still being useless for longer distances
- Rebalances mech power cages to have a more meaningful variety and adjusted costs, with even the smallest cage outclassing a hyper-capacity cell (since it's BIG and expensive).
- Rebalances Mech Fuel Cell to be very expensive but auto-charging (slowly) with a higher power ceiling, designed for the more power-hungry S2 and S4 mech weapons since it unlocks with the research for them.
- Tweaks the energy economy and stats of Frame weapons to reduce the dominance of the Ravager and improve the attractiveness of other Medium and Light Frame weapons.

## Why / Balance

You didn't ask for this, but it is something that bugged me, how the S2 mecha were effectively out-classed in practically every way by the S4s in everything but size, and how high-end mech teams would have a fairly hard upper limit on combat effectiveness.
The cheap Small Cage is basically doubled in capacity and gains a total of 25 MJ of EMP resistance, the Medium Cage increases its capacity by 70%, and the High-capacity Cage isn't changed too much while its costs go up substantially.
Mech Fuel Cell is made very high capacity (7200) and auto-charges slowly (30/sec, four minutes from empty to full, 12-second usage delay on recharging) at the cost of being very expensive (2 fissile uranium and 1 bluespace crystal among other costs, 'shielding' to keep it from being super radioactive). This extends the utility of peak-equipped mechs indefinitely as long as they don't need repairs, but still keeps a fairly tight budget on the use of very powerful anti-ship weapons in any given engagement. Probably very powerful on S1 mechs, but offset by the low durability and weaker weapons of S1 mechs.
Made the S2 mecha slightly-less-useless outside of their focus area while still keeping them less-than-optimal in terms of mobility. The Sabre is now fairly good at close-range EVA operations while still remaining vastly inferior to the Broadsword and Halberd in long-range space ops, and those two suffer slightly less if they need to enter a grid: I want to push it higher to 1.5 and be equivalent to the Marauder, but I don't think that's what the server wants.
During this PR, I discovered that the Ravager is pretty much the meta choice for Flails - it does amazing damage and is ridiculously economical compared to all other Frame weapons. It probably isn't enough, but this PR also tries to reduce some of its energy economy while improving the economy of the other Medium, Heavy, and Light Frame weapons though various means.

## How to test

Build/Spawn Broadswords, Flails, and their variants and run/fly around, load weapons with new cages/fuel cell and blam away.

## Media

n/a: Refactor

## Requirements

## Breaking changes

None.

**Changelog**

:cl:
- tweak: Walk speed of all S2 and S4 mecha increased to roughly half their Sprint speeds.
- tweak: Broadsword and Halberd Sprint speed increased from 1 to 1.33
- tweak: Sabre space speed doubled from 0.3 to 0.66
- tweak: Spyglass Sprint speed increased from 0.5 to 0.8
- tweak: Mace Sprint speed increased from 1.0 to 1.2
- tweak: Increased capacity of all Power Cages
- tweak: Increased Lathe costs of all Power Cages
- add: Various Power Cages can be printed directly in the Exofab instead of needing a merc/military fab.
- tweak: Mech Fuel Cell now costs Fissile Uranium (1) and Diamond (1) among other cost changes, but auto-regens to full within 4 minutes. 
- tweak: Balance pass on Armored Frame (S2/S4 mech) weapons - RAC-6 spreads slower, LAC-29 spreads slower and firing is cheaper, DMR-90 spreads slower and fires cheaper, ASM-55 fires cheaper, R3 has bigger boom and has actual but limited tracking, Metronome structure damage increased slightly and accuracy greatly increased, Ravager firing costs increased

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Power/powercells.yml
+++ b/Resources/Prototypes/Entities/Objects/Power/powercells.yml
@@ -322,9 +322,9 @@
       state: o2
       shader: unshaded
   - type: Battery
-    maxCharge: 2800
+    maxCharge: 2800 # Mono: 1400 >> 2800
     startingCharge: 2800
-  - type: EmpResistance
+  - type: EmpResistance # Mono also
     coefficient: 0.000112
 
 - type: entity
@@ -342,7 +342,7 @@
       state: o2
       shader: unshaded
   - type: Battery
-    maxCharge: 4600
+    maxCharge: 4600 # Mono: 2700 >> 4600
     startingCharge: 4600
 
 - type: entity
@@ -383,7 +383,7 @@
       state: o2
       shader: unshaded
   - type: Battery
-    maxCharge: 6400
+    maxCharge: 6400 # Mono: 6100 >> 6400
     startingCharge: 6400
 
 - type: entity
--- a/Resources/Prototypes/Entities/Objects/Power/powercells.yml
+++ b/Resources/Prototypes/Entities/Objects/Power/powercells.yml
@@ -311,7 +311,7 @@
   id: PowerCageSmall
   parent: BasePowerCage
   name: small-capacity power cage
-  description: A rechargeable power cage for big devices. This is the cheapest kind you can find.
+  description: A rechargeable power cage for big devices. This is the cheapest kind you can find, but it comes with some moderate EMP shielding.
   components:
   - type: Sprite
     sprite: Objects/Power/power_cages.rsi
@@ -324,6 +324,8 @@
   - type: Battery
     maxCharge: 2800
     startingCharge: 2800
+  - type: EmpResistance
+    coefficient: 0.25
 
 - type: entity
   id: PowerCageMedium
@@ -358,11 +360,13 @@
       state: o2
       shader: unshaded
   - type: Battery
-    maxCharge: 9001 # memes
-    startingCharge: 9001
+    maxCharge: 7200
+    startingCharge: 7200
   - type: BatterySelfRecharger
+    autoRechargePause: true
+    autoRechargePauseTime: 12 # Prevent continuous fire, but not a huge delay
     autoRecharge: true
-    autoRechargeRate: 30 # 5 minutes to fully recharge, 2.5x recharge rate of micro-reactor
+    autoRechargeRate: 30 # 4 minutes to fully recharge, 2.5x recharge rate of micro-reactor
 
 - type: entity
   id: PowerCageHigh
--- a/Resources/Prototypes/Entities/Objects/Power/powercells.yml
+++ b/Resources/Prototypes/Entities/Objects/Power/powercells.yml
@@ -364,7 +364,7 @@
     startingCharge: 7200
   - type: BatterySelfRecharger
     autoRechargePause: true
-    autoRechargePauseTime: 12 # Prevent continuous fire, but not a huge delay
+    autoRechargePauseTime: 30 # Prevent continuous fire, but not a huge delay
     autoRecharge: true
     autoRechargeRate: 30 # 4 minutes to fully recharge, 2.5x recharge rate of micro-reactor
 
--- a/Resources/Prototypes/Entities/Objects/Power/powercells.yml
+++ b/Resources/Prototypes/Entities/Objects/Power/powercells.yml
@@ -322,8 +322,8 @@
       state: o2
       shader: unshaded
   - type: Battery
-    maxCharge: 1400
-    startingCharge: 1400
+    maxCharge: 2800
+    startingCharge: 2800
 
 - type: entity
   id: PowerCageMedium
@@ -340,14 +340,14 @@
       state: o2
       shader: unshaded
   - type: Battery
-    maxCharge: 2700
-    startingCharge: 2700
+    maxCharge: 4600
+    startingCharge: 4600
 
 - type: entity
   id: PowerCageMech
   parent: BasePowerCage
   name: mech fuel cell
-  description: A rechargeable fuel cell for mechs.
+  description: An auto-recharging fuel cell for mechs.
   components:
   - type: Sprite
     sprite: Objects/Power/power_cages.rsi
@@ -358,8 +358,11 @@
       state: o2
       shader: unshaded
   - type: Battery
-    maxCharge: 6000
-    startingCharge: 6000
+    maxCharge: 9001 # memes
+    startingCharge: 9001
+  - type: BatterySelfRecharger
+    autoRecharge: true
+    autoRechargeRate: 30 # 5 minutes to fully recharge, 2.5x recharge rate of micro-reactor
 
 - type: entity
   id: PowerCageHigh
@@ -376,8 +379,8 @@
       state: o2
       shader: unshaded
   - type: Battery
-    maxCharge: 6200
-    startingCharge: 6200
+    maxCharge: 6400
+    startingCharge: 6400
 
 - type: entity
   id: PowerCageSmallEmpty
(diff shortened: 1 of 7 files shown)


dge21:
finally, good mechs other than the durand

Resources/Prototypes/_Mono/Recipes/Lathes/mech_parts.yml
```diff
@@ -194,10 +194,14 @@
...
-    Steel: 600
-    Glass: 800
-    Plastic: 400
+    Plasteel: 400
+    Glass: 400
+    Plastic: 200
+    Silver: 100
+    Plasma: 200
     Gold: 100
+    UraniumFissile: 50
```
Helicaste: Cost is amount / 100, this would be 0.5 Fissile and 0.05 Bluespace Crystal.

For reference a Microreactor costs 0.5 Fissile.

You could increase to 2 Fissile (200) and it would be _a_ price for the provided utility, given that cost is essentially ~300 Uranium + ~300 Plasma ore.

Resources/Prototypes/Entities/Objects/Power/powercells.yml
```diff
@@ -358,8 +358,11 @@
       state: o2
       shader: unshaded
   - type: Battery
-    maxCharge: 6000
-    startingCharge: 6000
+    maxCharge: 9001 # memes
```
Helicaste: Just for clarity to the mech balancing, a Ravager Plasma Cannon costs 70 energy per shot, so this would be able to fire it 128 times successively.

Helicaste (commented):
See inline comments;

I might also recommend two things;
-Making the mech reactor have a recharge offset timer (See Apollo laser for an example) for a higher recharge rate, it rewards pulling back from combat to recharge your battery instead of having a constant flat rate.

-Change the function of the Small Power Cage to be a Combat Power Cage with EMP resistance but the same relatively low capacity seen in the Combat Power Cell. All Cages except Mech are unlocked at the same time so as-is there's essentially no reason to use the Small one.

Resources/Prototypes/_Mono/Recipes/Lathes/mech_parts.yml
Kiithnaras: In that case, I'm pretty strongly tempted to _increase_ the costs of all the cages by a substantial margin since they're supposed to be fairly large and hefty items.

Resources/Prototypes/Entities/Objects/Power/powercells.yml
Kiithnaras: That's fair, the Ravager is pretty nasty and power cheap compared to some of the other big guns. Lowering the cap a bit to 7200 and adding a 12 second delay to the auto charge feels pretty good to me.

Kiithnaras:
I do want to keep the Small Cage at a substantial advantage to the Hyper Power Cell, but giving it a low amount of EMP resist makes sense to me (which I need to adjust again because I didn't test and compare values)

Resources/Prototypes/Entities/Objects/Power/powercells.yml
Kiithnaras: At only 70 charge for 400 structural with mild explosive radius compared to the Metronome at 100 charge for only 200 structural and Bazooka at 500 charge for 500 structural (but a big boom), it feels like the Ravager is out of line with the other Heavy mecha guns. Metro is definitely a sniper, Bazooka is a close-in bomb tosser, but the ravager just outclasses both in pretty much every regard except range and velocity on the metronome. I might want to increase the Ravager's charge cost to 120-150, increase the metro's damage to 250, and reduce the Bazooka's charge cost to 250 to better balance them, even with the recognition that the ravager has to chew through shields and lacks precision at ranges the frame can actually dodge return fire.

Helicaste: Bazooka could use a fairly significant structural damage increase to trade off its very low speed.

Metronome could mostly use a massive spread decrease imo.

Helicaste:
> I do want to keep the Small Cage at a substantial advantage to the Hyper Power Cell, but giving it a low amount of EMP resist makes sense to me (which I need to adjust again because I didn't test and compare values)

I’d meant similar balancing formula compared to the combat cell, hiking its price up. People making S2/S4 mechs have plenty of materials you can tax.

NazrinNya (changes requested):
probably make mech fuel cell cheaper (1 uranium fissle is enough), also why bluespace? bluhspace is very specific material and has no reason to be there.

also probably make low-high cages available in exofab.

Kiithnaras:
> probably make mech fuel cell cheaper (1 uranium fissle is enough), also why bluespace? bluhspace is very specific material and has no reason to be there.
> 
> also probably make low-high cages available in exofab.

Thinking behind bluespace crystal is that it is a sufficiently rare and expensive material to discourage over-reliance and also does lorewise wibbly stuff with spacetime. Can easily replace with Diamond, though, no biggie.



> > I do want to keep the Small Cage at a substantial advantage to the Hyper Power Cell, but giving it a low amount of EMP resist makes sense to me (which I need to adjust again because I didn't test and compare values)
> 
> I’d meant similar balancing formula compared to the combat cell, hiking its price up. People making S2/S4 mechs have plenty of materials you can tax.

Fair, the intention is for the power cages to be more affordable for S1 mechs as well though. A bit of diamond and uranium probably would work just fine to that effect though.

Helicaste:
> > probably make mech fuel cell cheaper (1 uranium fissle is enough), also why bluespace? bluhspace is very specific material and has no reason to be there.
> > also probably make low-high cages available in exofab.
> 
> Thinking behind bluespace crystal is that it is a sufficiently rare and expensive material to discourage over-reliance and also does lorewise wibbly stuff with spacetime. Can easily replace with Diamond, though, no biggie.
> 
> > > I do want to keep the Small Cage at a substantial advantage to the Hyper Power Cell, but giving it a low amount of EMP resist makes sense to me (which I need to adjust again because I didn't test and compare values)
> > 
> > 
> > I’d meant similar balancing formula compared to the combat cell, hiking its price up. People making S2/S4 mechs have plenty of materials you can tax.
> 
> Fair, the intention is for the power cages to be more affordable for S1 mechs as well though. A bit of diamond and uranium probably would work just fine to that effect though.

Again, anyone making these mechs will already have a large supply of Materials, especially for bluespace because there are not many things that require it. The only limiting factor of the Reactor is Fissile, which costing only One fissile will still cost ~150 Uranium and Plasma Ore apiece.

Kiithnaras:
> Again, anyone making these mechs will already have a large supply of Materials, especially for bluespace because there are not many things that require it. The only limiting factor of the Reactor is Fissile, which costing only One fissile will still cost ~150 Uranium and Plasma Ore apiece.

I was mainly referring to the non-Reactor cages being "affordable" for the sake of S1 exos. The Fuel Cell should absolutely need Fissiles, and I adjusted those based on the guidance from NazrinNya. Regardless, new changes submitted for review and I at least think they adhere to the feedback here, lmk if I'm missing something.

Still think the heavy guns need some balancing, but I'll save that for another commit.

Kiithnaras (commented):
Swapped BSC for Diamond in Fuel Cell cost, reduced FissileU cost to 1 sheet as requested, further adjustments applied in later commits based on discussions.

NazrinNya (approved):

PR #3277, opened 2026-02-23
Decreases AK570 Burst Cooldown

Redrover1760:
## About the PR

Decreases AK570 burst cooldown, increasing the overall rate of fire of the AK570.

## Why / Balance

Currently in testing, the Dravon outperforms the AK570 by every metric. it has more DPS against a target, breaks walls faster, has more range and projectile speed, the only thing the AK570 has going for it is volume of fire, which isn't particularly significant.

This should give the AK570 more strength to compete.

either that or nerf 90mm or something I dunno.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: AK570 burst cooldown decreased (firerate increase)

### DIFF
--- a/Resources/Prototypes/_Mono/Entities/SpaceArtillery/SpaceArtillery/Kinetic/standardmedium_kinetic.yml
+++ b/Resources/Prototypes/_Mono/Entities/SpaceArtillery/SpaceArtillery/Kinetic/standardmedium_kinetic.yml
@@ -23,7 +23,7 @@
     projectileSpeed: 170
     minAngle: 0
     maxAngle: 3
-    burstCooldown: 0.4
+    burstCooldown: 0.3
     burstFireRate: 10
     shotsPerBurst: 2
     selectedMode: Burst


OneZerooo0:
How fast does it pen now? If it’s too fast we should nerf 90mm instead

tonotom1:
> How fast does it pen now? If it’s too fast we should nerf 90mm instead

shitbrick.yml abuse tests are in

Sustained fire for 60 seconds, each test

Count tiles destroyed and not penetration depth

AK570 @ 0.4 refire
image

AK570 @ 0.3 refire
image

Dravon
image

Bofors
image


In real scenarios where both targets are moving, the AK570 will perform significantly worse vs armor due to its lower projectile velocity and lower damage density per projectile so this buff wouldn't be out of the question I feel. It'd also be possible to increase general effectiveness without raw DPS by increasing projectile velocity if you feel that cranking its fire rate is too goofy.

Redrover1760:
Thank you for screenshotting the testing ye. That's basically exactly what I found.

It could also be reasonable to or nerf damage of either.

OneZerooo0:
I think dravons should be slightly nerfed instead

Redrover1760:
yeah a single dravon probably shouldn't outperform a bofor

tonotom1:
> yeah a single dravon probably shouldn't outperform a bofor

It doesn't though, it's 11 tiles destroyed vs 18 tiles. Both cost 6 processing power to field. You need to count tiles destroyed and not penetration depth since the likelihood of hitting the same place like in a controlled test is effectively impossible due to movement and innate spread over true firing distances. Regardless, there's ample room to reduce its effectiveness in comparison to the AK570.

PR #3285, opened 2026-02-24
Hullmods 

NazrinNya:
## About the PR

Adds new `GridModifier` System along with its own prototype and component.
Currently its capable of replacing each entity on the grid.
## Why / Balance

Capabilities of generating derelict ships with bunch of damage mods, or advanced ships with good mods.
## How to test

Load in
Spawn some shuttles
Apply hullmods to them using `addgridmodifier`
## Media

Arkansaw with ReinforcedHull Hullmod
изображение

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added Hullmods, currently admin-only.


Content.Server/_Mono/Grid/Modifiers/EntityReplace.cs
```diff
@@ -0,0 +1,53 @@
+using Content.Shared.Whitelist;
+using JetBrains.Annotations;
+using Robust.Shared.Prototypes;
+
+namespace Content.Server._Mono.Grid.Modifiers;
+
+[UsedImplicitly]
+public sealed partial class EntityReplace : GridModifier
```
Ilya246: should probably be called GridModEntityReplace

Content.Shared/_Mono/Grid/GridModifier.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Content.Server._Mono.Grid.Modifiers;
+
+[ImplicitDataDefinitionForInheritors]
+[MeansImplicitUse]
+public abstract partial class GridModifier
+{
+    protected string _id => GetType().Name;
+
+    [DataField]
+    public string Comp = "Transform";
```
Ilya246, line 12: i think there is some way to enforce this being a comp name, maybe see what ComponentRegistry does

Content.Server/_Mono/Grid/GridModifierSystem.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                var comp = _factory.GetComponent(modifier.Comp);
+                var ents = new HashSet<Entity<IComponent>>();
+
+                GetGridEntities(uid, ents, comp.GetType());
+
+                foreach (var ent in ents)
+                {
+                    modifier.Modify(ent, _metaQuery.Get(ent), _xformQuery.Get(ent), EntityManager);
+                }
```
Ilya246: this should ideally be handled by the grid modifier, all this system should do is call the modifier with the grid as arg
also don't use xform query or metadata query, use Transform() and MetaData()

Content.Client/Entry/EntryPoint.cs
```diff
@@ -126,6 +126,8 @@ public override void Init()
             _prototypeManager.RegisterIgnore("nukeopsRole");
             _prototypeManager.RegisterIgnore("ghostRoleRaffleDecider");
             _prototypeManager.RegisterIgnore("pointOfInterest"); // Frontier: worldgen-related, server-only
+            _prototypeManager.RegisterIgnore("gridModifier"); // mono
```
Ilya246: just declare it in shared even if the logic is on server
also makes it easier if we want to e.g. put it in UI later

Content.Server/_Mono/Grid/Modifiers/EntityReplace.cs
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+}
+
+[DataDefinition]
+[Serializable]
+public sealed partial class ReplaceData
+{
+    [DataField]
+    public EntityWhitelist? Whitelist;
```
Ilya246: do an alias getter/setter variable for Whitelist.Tags (or maybe 2 aliases, Tags if you want a list and Tag if you want one tag) so you can do tag whitelist without bloat

Ilya246 (changes requested):

Content.Shared/_Mono/Grid/SharedGridModifierSystem.cs
Ilya246, line 1: a shared system isn't needed here

Ilya246 (approved):

PR #3288, opened 2026-02-25
Improves AME, Nerfs ZPinch

Kiithnaras:
## About the PR

Added a positive exponential scalar for power output for AME core counts; bigger cores produce more power for the same amount of fuel, and AMEs produce about the same amount of power per core at 2x fuel per core regardless of size.
Enlarged AME jars to hold 1600 fuel instead of 1250.
Reduced ZPinch fuel efficiency to run just under two hours (105 minutes) on roundstart fuel.
Introduced logic checks to prevent the AME controller from producing power without any cores and from registering negative-infinity power at 0 injection.

## Why / Balance

Z-Pinch very too good. AME big and not as good. Made this better. Big AME very much power. Z-Pinch need refueling during round now. Seemed silly to be flying with a giant explosive power plant that might only be marginally better than a pinch reactor or two.

## How to test

Fly ships with AMEs and fly others with Pinches.

## Media

n/a

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Prevented AME controller from producing risk-free power with no cores attached
- fix: Prevented AME controller from registering negative-infinity power at zero injection
- tweak: Reduced Z-Pinch fuel efficiency
- tweak: Increased AME Jar fuel amount
- tweak: Increased AME core-based power scaling - larger AMEs produce power more efficiently at low injection rates and have about the same power efficiency at max stable fuel rate.

### DIFF
--- a/Content.Server/Ame/AmeNodeGroup.cs
+++ b/Content.Server/Ame/AmeNodeGroup.cs
@@ -180,7 +180,7 @@ public sealed class AmeNodeGroup : BaseNodeGroup
         if (cores > 0 && fuel > 0)
             return 200000f * MathF.Log10(fuel * fuel) * MathF.Pow( 1.11111f, cores - 1); // Mono - Added positive returns from extra cores. Running AME at 2 fuel per core yields roughly the same power per core at all sizes
         else
-            return 0
+            return 0;
     }
 
     public int GetTotalStability()
--- a/Content.Server/Ame/AmeNodeGroup.cs
+++ b/Content.Server/Ame/AmeNodeGroup.cs
@@ -177,7 +177,10 @@ public sealed class AmeNodeGroup : BaseNodeGroup
         // At 18+ cores and 2 inject, the power produced is less than 0, the Max ensures the AME can never produce "negative" power.
         // return MathF.Max(200000f * MathF.Log10(2 * fuel * MathF.Pow(cores, (float)-0.5)), 0); // Frontier: preferring old calculation for now
         // return 200000f * MathF.Log10(fuel * fuel) * MathF.Pow(0.75f, cores - 1); // Frontier: preferring old calculation for now
-        return 200000f * MathF.Log10(fuel * fuel); // Mono - Removed diminishing returns from AME core count.
+        if (cores > 0 && fuel > 0)
+            return 200000f * MathF.Log10(fuel * fuel) * MathF.Pow( 1.11111f, cores - 1); // Mono - Added positive returns from extra cores. Running AME at 2 fuel per core yields roughly the same power per core at all sizes
+        else
+            return 0
     }
 
     public int GetTotalStability()
--- a/Content.Server/Ame/AmeNodeGroup.cs
+++ b/Content.Server/Ame/AmeNodeGroup.cs
@@ -177,10 +177,9 @@ public sealed class AmeNodeGroup : BaseNodeGroup
         // At 18+ cores and 2 inject, the power produced is less than 0, the Max ensures the AME can never produce "negative" power.
         // return MathF.Max(200000f * MathF.Log10(2 * fuel * MathF.Pow(cores, (float)-0.5)), 0); // Frontier: preferring old calculation for now
         // return 200000f * MathF.Log10(fuel * fuel) * MathF.Pow(0.75f, cores - 1); // Frontier: preferring old calculation for now
-        if (cores > 0 && fuel > 0)
+        if (cores > 0 && fuel > 0) // Mono - default zero power unless conditions are met
             return 200000f * MathF.Log10(fuel * fuel) * MathF.Pow( 1.11111f, cores - 1); // Mono - Added positive returns from extra cores. Running AME at 2 fuel per core yields roughly the same power per core at all sizes
-        else
-            return 0;
+        return 0;
     }
 
     public int GetTotalStability()
--- a/Content.Shared/Ame/Components/AmeFuelContainerComponent.cs
+++ b/Content.Shared/Ame/Components/AmeFuelContainerComponent.cs
@@ -9,11 +9,11 @@ public sealed partial class AmeFuelContainerComponent : Component
     /// The amount of fuel in the container.
     /// </summary>
     [DataField, ViewVariables(VVAccess.ReadWrite), AutoNetworkedField]
-    public int FuelAmount = 1250; // mono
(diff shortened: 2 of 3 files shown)


Content.Server/Ame/AmeNodeGroup.cs
```diff
@@ -177,7 +177,10 @@ public float CalculatePower(int fuel, int cores)
         // At 18+ cores and 2 inject, the power produced is less than 0, the Max ensures the AME can never produce "negative" power.
         // return MathF.Max(200000f * MathF.Log10(2 * fuel * MathF.Pow(cores, (float)-0.5)), 0); // Frontier: preferring old calculation for now
         // return 200000f * MathF.Log10(fuel * fuel) * MathF.Pow(0.75f, cores - 1); // Frontier: preferring old calculation for now
-        return 200000f * MathF.Log10(fuel * fuel); // Mono - Removed diminishing returns from AME core count.
+        if (cores > 0 && fuel > 0)
+            return 200000f * MathF.Log10(fuel * fuel) * MathF.Pow( 1.11111f, cores - 1); // Mono - Added positive returns from extra cores. Running AME at 2 fuel per core yields roughly the same power per core at all sizes
+        else
```
Ilya246: there's a return under the if you don't need an else

Content.Server/Ame/AmeNodeGroup.cs
```diff
@@ -177,7 +177,10 @@ public float CalculatePower(int fuel, int cores)
         // At 18+ cores and 2 inject, the power produced is less than 0, the Max ensures the AME can never produce "negative" power.
         // return MathF.Max(200000f * MathF.Log10(2 * fuel * MathF.Pow(cores, (float)-0.5)), 0); // Frontier: preferring old calculation for now
         // return 200000f * MathF.Log10(fuel * fuel) * MathF.Pow(0.75f, cores - 1); // Frontier: preferring old calculation for now
-        return 200000f * MathF.Log10(fuel * fuel); // Mono - Removed diminishing returns from AME core count.
+        if (cores > 0 && fuel > 0)
+            return 200000f * MathF.Log10(fuel * fuel) * MathF.Pow( 1.11111f, cores - 1); // Mono - Added positive returns from extra cores. Running AME at 2 fuel per core yields roughly the same power per core at all sizes
```
Ilya246, line 181: just change this to 60k * fuel?

Content.Shared/Ame/Components/AmeFuelContainerComponent.cs
```diff
@@ -9,11 +9,11 @@ public sealed partial class AmeFuelContainerComponent : Component
...
     [DataField, ViewVariables(VVAccess.ReadWrite), AutoNetworkedField]
-    public int FuelAmount = 1250; // mono
+    public int FuelAmount = 2000; // mono
 
     /// <summary>
     /// The maximum fuel capacity of the container.
     /// </summary>
     [DataField, ViewVariables(VVAccess.ReadWrite), AutoNetworkedField]
-    public int FuelCapacity = 1250; // mono
+    public int FuelCapacity = 2000; // mono
```
Ilya246: no

Resources/Prototypes/_Mono/Entities/Structures/Power/Generators/generators.yml
```diff
@@ -193,7 +193,7 @@
     maxTargetPower: 160000
     targetPower: 105000
     optimalPower: 105000
-    optimalBurnRate: 0.001
+    optimalBurnRate: 0.00333
```
Ilya246: just do 0.004

Ilya246 (changes requested):

Content.Shared/Ame/Components/AmeFuelContainerComponent.cs
Kiithnaras: Curious as to why. At minimum rate, 1250 lasts just under two hours (1:44). For an AME at a rate of 6 per injection every 10 seconds (ideally a 3-core or bigger), this only lasts 30 minutes. If that's an acceptable balance point, no big deal.
2000 _is_ probably too much upon reflection. Would 1600 be acceptable, lasting just under 45 minutes at 6 or 2:13 at 2? Or is this just a hard nope and don't touch it?

Content.Server/Ame/AmeNodeGroup.cs
Kiithnaras: That could work but it would generate a very powerful linear power output. The square-logarithm fuel rate element makes higher fuel rates give diminishing returns. Adding a competing expanding-return core-based element gives _some_ benefit to larger  AMEs (takes more space, makes bigger boom, easier to hit) even at reduced fuel rate and competes with the diminishing one enough to result in a roughly-linear size-power relationship at maximum stable fuel rate.
My thought doing it this way was to provide a benefit for larger singular AME arrays over multiple independent single-core AMEs without making them explicitly more powerful than multiple single-cores.
Lemme know what you think.

Content.Server/Ame/AmeNodeGroup.cs
Kiithnaras: Fair, seems like a stylistic difference more than anything. I find I prefer explicitness and clarity, but I'm not married to my particulars.

Content.Shared/Ame/Components/AmeFuelContainerComponent.cs
Ilya246: okay sure 1600

Content.Server/Ame/AmeNodeGroup.cs
Ilya246: actually you know what sure

Ilya246:
image
1 here is 200k, x is core count (max safe injection)

Ilya246 (approved):

Kiithnaras:
desmos-graph
This is a comparison I used for visualization. Vertical is power in kW, horizontal is fuel rate.
- Blue is the 60*fuel relationship.
- Green is the fuel-power curve as proposed for a single-core AME (and the current power curve for all AMEs).
- Purple is the fuel-power curve as proposed for a four-core AME.
- Black is the maximum safe injection relationship with the proposed model. It does diverge from linear at very high core counts, but I don't think I've seen any grid/ship with more than a six-core AME.

Ilya246:
i think being able to build a 50 core giga AME that can power a city is fun

PR #3327, opened 2026-03-02
Fix Dynamic Role Timer

Ilya246:
## About the PR
it ignores your cvars and just sets them to true
also fixes debug strength thrusters because those were broken

no CL


Redrover1760:
so real

Redrover1760 (approved):
if tested goida

Ilya246:
not tested but should work, this is relevant for multiplayer localhosts

mikusssssss:
WTF!!! stealth change!!! you modified unspawnable item???? repoban

PR #3336, opened 2026-03-03
Makes T2 Hardsuits Usable

Redrover1760:
## About the PR

Reworks the M86 into a Heat resistant suit first, with less blunt slash and pierce.

Gives the Jackal an edge in heat resistance over the CV-32.

Gives the Viper a 5% speed boost.

## Why / Balance

As discussed before during the initial implementation, T2 suits should be viable. T2 suits should grant options the T3 ones don't, even if they are overall less strong. T3 suits may be uber tanky jugsuits or fast speedsuits with good armor, but T2 suits should occupy a strong middleground between armor and mobility, or have some kind of usecase that leaves them relevant and useful gear, even with T3 suits unlocked.

they are currently not so. T3 suits outperform T2 by every metric. T1 suits even outperform T2 by most metrics. T2 suits are essentially useless. There is no reason to use one pretty much and every reason to immediately rush to T3 suits as the T2 ones are just that much horrendously worse. God forbid you don't acquire the magic T3 suits as you basically die. M86 is however a much much worse example of this and I'm disappointed to think that the unusable 20% slowdown suit with 20% heat resistance was not tweaked in 4 months whatsoever.

M86: Gives TSFMC a vitally needed option against incen ammo spam. Makes the M86 have a usecase period instead of dying to any amount of heat weaponry with 20% heat resist for no appreciatable tradeoff.

Jackel: Makes the Jackal more heat resistant than the CV-32, useful and gives it a permanent usecase against heat-based weaponry over the T3 suit. 

Viper: Gives the Viper a use case at all with a minor speed buff.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: T2 faction suits buffed, specifically the M86, Jackal, and Viper.
- tweak: M86 is now a heat-resistant suit first with 65% Heat resistance and less other resistances.
- tweak: Jackal now has 60% heat resistance, making have a usecase over the CV-32 against incen weaponry while not forfeiting all resistances to other weapons by using the CV-53.
- tweak: Viper suit now has 5% speed bonus. This makes it have a niche at all for a T2 suit.


freturt:
Imho you should make then m86 at least more durable against ballistics or make it' more mobile because 20% slowness for 40% pierce is a lot, it kinda looks like that viper or even m82c hardsuits are better variants to use.
Or just make m86 tier 3 hardsuit and buff all of its resists so TSFMc finally gets their version of jugsuit ngl 😃

Mrrpip:
Balance aside, I think if you're trading armor for heat on the M86 MK.3, the description should be adjusted.

Redrover1760:
> Imho you should make then m86 at least more durable against ballistics or make it' more mobile because 20% slowness for 40% pierce is a lot, it kinda looks like that viper or even m82c hardsuits are better variants to use. Or just make m86 tier 3 hardsuit and buff all of its resists so TSFMc finally gets their version of jugsuit ngl 😃

Its 45% pierce. But I could consider increasing its mobility to 15% slowdown yes.

OneZerooo0 (changes requested):
Make T1 suits heat resistant instead for now, M86 being pierce resistant is fine

PR #3364, opened 2026-03-05
Chimera Drip

Ilya246:
## About the PR
gives chimera a bunch of clothing slots
however they get no effects from clothing except obscuring name

## Why / Balance
i am legitmate person not chimera trust ignore red skin

## Media
image

the chimera arms still poke out but i'm not sure how to fix that so i say feature

**Changelog**

:cl:
- add: Chimera can now wear clothing, but get no effects from it.

### DIFF
--- a/Content.Shared/Inventory/InventoryComponent.cs
+++ b/Content.Shared/Inventory/InventoryComponent.cs
@@ -33,6 +33,12 @@ public sealed partial class InventoryComponent : Component
     /// </summary>
     [DataField]
     public Dictionary<string, DisplacementData> MaleDisplacements = new();
+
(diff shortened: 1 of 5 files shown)


Mrrpip:
I feel like this is a bad idea.

Even if it is fairly obvious if you check for the spikes, KOS entities should be noticeable at a glance.

Ilya246:
it is noticeable the arms poke out a lot

NazrinNya (changes requested):
head slot should have offset like borgs do

Ilya246:
fixed
image

NazrinNya (approved):

PR #3383, opened 2026-03-06
FTL Drive Changes

OneZerooo0:
## About the PR

Changes existing FTL drives into various phase drives, which allow for quick ducks out of combat - fast activation but low range FTL drives that allow for tactical repositioning. This also means you can't teleport across the map anymore.
## Why / Balance

We have higher max speeds and relatively close-by POIs that allow for "normal" sublight travel to be viable, and point to point instant travel has multiple problems - namely, quick/easy escapes from combat and inability for others to intercept people who are traveling.
Additionally, a tactical dodge should add more complexity to combat - choose when to dodge incoming fire, and reposition occasionally in disadvantageous positions.
## How to test

use ftl
## Media

https://github.com/user-attachments/assets/eb971b23-72ef-46c6-957c-5ac5a0743ae6

## Requirements

## Breaking changes

**Changelog**

:cl: Onezero0
- remove: Removed long-distance FTL.
- tweak: All FTL drives have become tactical phase drives, and allow for quick combat repositioning.


Resources/Prototypes/_Mono/Entities/Structures/Machines/ftldrive.yml
```diff
@@ -116,7 +119,10 @@
       shader: unshaded
       color: "#E6173E"
   - type: FTLDrive
-    range: 10000
+    range: 250
```
UnicornOnLSD: shouldn't that be 500 or something ? 
If not it's just a roll with double the cooldown and triple the hyperspace time.

OneZerooo0: tweaked

DieselMohawkTheSequel:
The issue for this is that the FTL menu is VERY CLUNKY.
Any situation where you're the gunner and the pilot is clunky and makes it nearly useless, then the alternative is to have a gunner and a pilot which will piss off the gunner when they suddenly teleport.

DieselMohawkTheSequel:
It would make more sense if it were used via the NAV menu, since that's the menu where you see ships and projectiles.

Kiithnaras:
Shuttles on drive cooldown also have their Parking flight modes disabled, as a note. Not sure if you want to change this or reduce cooldowns.
I don't mind the idea of cutting ranges, particularly cutting them dramatically, but reducing them so much and requiring players to fly long distances and dodging rocks and wrecks might get tedious with as large and cumbersome as many shuttles can be.

I'd recommend instead cutting the drive distances by a factor of 5 to 10:
1000m short-hop with a cooldown of 30-40 for the Roll5
500m hops with a 15-20 sec cooldown for the Dodge2
2000m medium hops with a cooldown of 60 seconds for the Sprint15
ECES-60 (ECES-30?) maintains its range dominance with ca. 6km-8km range while increasing its cooldown to 90 seconds but shortening its base transit time to 30 seconds and increasing its spin-up time to 6 seconds.
I'd also encourage restoring a minimum range of 100m to the phase shift lockout.

This ensures that the smaller "combat" drives can't pull you out of a fight entirely, and escaping is easier but not a guarantee since enemies will have to maintain a much closer proximity, putting themselves at risk in turn. Much easier to escape ships trying to snipe with missiles and mass drivers compared to brawlers, and puts brawlers at risk if they don't intend to ram.

Zethine07:
This change would be perfectly fine if it was applied to combat grade vessels only. As it stands, civilian vessels are affected too, when they arent even meant for or made with combat in mind.

DieselMohawkTheSequel:
> This change would be perfectly fine if it was applied to combat grade vessels only. As it stands, civilian vessels are affected too, when they arent even meant for or made with combat in mind.

This will ruin gameplay for people who need medics, the travel time will grow exponentially and the dead people will pay the price.

DieselMohawkTheSequel:
I meant that in support of the comment, not arguing against it, civ shuttles should be capable of fast and safe travel.

Kiithnaras:
You should also remember that weapons and shields are disabled while the drives are on cooldown. For larger ships that have relied on the CTLA-160, four minutes of no shields or guns would be devastating.

AthrunZala33092:
If this is meant for FTLing and positioning at close range, it's EXTREMELY pointless, and with the time it takes to FTL + arrival time it would be faster to just fly the 250-500m distance. It also makes ships vulnerable with the FTL cooldown taking shields and weapons offline. Large ships become especially vulnerable for a solid 40 seconds or so not being able to shoot back and no shields up. This change honestly makes FTL for the new intended function more detrimental to use. It also makes reinforcements now completely impossible if we're keeping the normal FTL cooldown.

As everyone else said, this also affects: anyone trying to get ANYWHERE in the galaxy, such as the rim, medics rescuing people with as low as rot is. Honestly this effects civilians the MOST, and makes the combat application for faction ships kind of useless. We might as well just not have FTL drive at all and carrion fly the way. It'd be faster and we'd at least have shields up.

Redrover1760:
This PR is to only be considered for merge after asteroid belts are in place to make sublight flight much less frustrating on that note

and cooldowns need to be made not deactive weapons yeah

Feedback thread created to discuss, see https://discord.com/channels/1329292619834069034/1479629133159076021 in #Feedback

OneZerooo0:
PR is on hold due to feedback + need for other content to necessitate its existence

Ilya246:
autopilot PR is up for this to be mergeable

PR #3394, opened 2026-03-08
Autopilot

Ilya246:
## About the PR
adds autopilot button next to ftl button
works same way ftl does but instead of ftl it makes your ship drive itself to destination
auto-cancel on using any manual controls
popup+sound on finish
uses shipHTN

## Why / Balance
part of planned FTL rework and just nice to have

## How to test
buy ship
autopilot a long distance
watch it dodge asteroids

**Changelog**

:cl:
- add: Shuttle consoles now come with autopilot in the map tab.


OneZerooo0 (approved):
goidafail
only thing that I feel like it should have is an indicator on the FTL screen of where autopilot is currently trying to go to (destination marker) but that can be done later

OneZerooo0:
we have enough testing of ship HTN in general that I think its safe to just merge this, naz should review though

PR #3413, opened 2026-03-09
Fix the Tests

NazrinNya:
## About the PR

tests now ignore all entities with spawner category, no need to test them.


NazrinNya:
ez

PR #3427, opened 2026-03-10
Fang Buff, More Projectile Speed/Range

Redrover1760:
## About the PR

Buffs the Fang projectiles from 170 m/s to 240 m/s

Increased firerate slightly.

Decreased inaccuracy.

## Why / Balance

Fang was still 170 m/s, which is what the AKS had, despite being switched over to 57mm while the Fang still being 90mm. The Dravon has 250 m/s in total.

Being 90mm ammo this reduced its range a ton which doesn't seem ideal versus standard dravons and the like. This was an oversight from the 57mm/90mm switcharoo.

Now is slightly less speed, more firerate, more inaccurate dravon type deal.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Fang projectile speed buffed. Old: 170 m/s New: 240 m/s. 
- tweak: Fang range increased. Old: 714m New: 1008m.
- tweak: Fang fires slightly faster, less inaccurate.


OneZerooo0:
I think 90mm is enough to justify current stats, though someone with more ship combat experience should weigh in

NazrinNya (approved):
good because they had roughly same/worse DPS than dravons.

PR #3443, opened 2026-03-14
Massive Stealth-Suit Buff

enkidienne:
## About the PR
Cybersun Stealth suit no longer loose stealth on movement. 

## Why / Balance
Stealth suit does not see any use because it, honestly is not very good. This buffs their bottom baseline by a lot.

## How to test
spawn the suit-move-it don't go down.

**Changelog**

- tweak: sets the movementVisibilityRate of the Cybersun Stealth suit to 0.

### DIFF
--- a/Resources/Prototypes/_Goobstation/Entities/Clothing/Head/hardsuit-helmets.yml
+++ b/Resources/Prototypes/_Goobstation/Entities/Clothing/Head/hardsuit-helmets.yml
@@ -63,6 +63,6 @@
       minVisibility: 0.15 # Mono: Changed to be not fully invisible
       maxVisibility: 0.6 # Mono: Changed to be partially invisible
     - type: StealthOnMove
-      passiveVisibilityRate: 0
-      movementVisibilityRate: 0.4
+      passiveVisibilityRate: -0.4
+      movementVisibilityRate: 0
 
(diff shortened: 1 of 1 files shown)


Helicaste:
I'd recommend some combination of;
-Add a small lighting effect similar to the cobra gland.
-Improve its movement visibility rate to be 0.1 or 0.05, instead of 0.0 (10 or 20 seconds of movement to decloak fully, iirc)
-Improve its passive visibility rate to enter stealth faster, -0.6 or -0.8
-Change its max visibility to be 0.4 or 0.3, or go the other way and make it 0.6/0.8.

If you wanted to be a bit spicier you could also make it not have pressure protection for better cloaking so it's a **real** stealth suit but only in atmosphere. Good for raiding, bad for sitting in space completely undetectable with voidboots.

Dealer's choice for what kind of suit you're trying to make I suppose. I'd prefer to "see" something that doesn't further incentivize ASR-type gameplay and un-reactable deaths.

Arcticular:
Pretty sure we can just move it to jugsuit tier in experimental and ramp up the price to be an equivalent.

mikusssssss:
yeah let’s just make the stealth suit turn you into the hidden from that one gmod gamemode bro

burrrkeye:
another content that heavily punishes a newbie by simply existing while only causes a toe stracthing levels of effort for an old player to handle

NazrinNya:
too powerful and unbalanced if you think about it for a second.

however, I can make turrets ignore stealthed mobs so it will have a use.

PR #3508, opened 2026-03-21
[Testmerge End] Revert Movable POIs

Ilya246:
merge and publish during round 1375 and before 1376

**Changelog**

:cl:
- remove: Movable POIs


Redrover1760 (changes requested):
anchor fixes are independent of mobile POIs (and don't allow POIs to be mobile), they don't need to be reverted

Ilya246:
ok

PR #3515, opened 2026-03-22
Expands Global Passive Healing & Buff Vox Space Resistances

Kiithnaras:
## About the PR

Rebalanced species passive healing to be broader, slightly slower, and with more species-specific quirks for Vox and Slimes.
Passive healing now functions as long as the mob is Alive (not Critical and definitely not Dead), regardless of thresholds (up to 200 technically, but must still be Alive). Most species have a low level of Heat, Cold, and Brute healing, and a very low level of Shock, Radiation, and Poison recovery.
Slimes recover much better in general, even adding a little bit of Cellular regeneration, at the cost of greatly reduced cold damage recovery.
Vox retain their enhanced poison recovery (albeit slower with no cap) and gain traits similar to slimes with enhanced radiation and asphyxiation recovery while also reducing their Barotrauma damage and lowering their speed loss thresholds.

## Why / Balance

This mimics the theme of upstream offmed and passive healing while tuning it for Monolith's gameplay. With the rightful sunset of the Platelet Factories Meta, every Bob, Geoff, and Susan are going to be chugging meds like they're candy with no reasonable source of passive healing, especially when most guns and weapons exceed that passive healing threshold already.
This will not return the previous status of PF players being "PvE Gods," not that they ever were before, nor will it expand that status to everyone; the rate of healing, even for slimes, is too slow, and it only functions when not in Crit. It will ensure that medical is not flooded with players demanding treatment for every bump, scrape, cut, brief glimpse at the reactor, stubbed toe, or burnt finger, but able to serve players actually looking for and needing real medical care and wanting actual roleplay. Healing in the short term for rapid return to combat will still require at least basic first aid if not dedicated meds.
This also gives a little more mechanical flavor to Vox with low pressure/temperature tolerances in ways different from Hydrakin, keeping with the lore/concept of Vox being bioengineered for space scrapping.

## Media

n/a

## Requirements

## How to test

Play as your favorite organic species (Sorry IPCs), get injured.

## Breaking changes

**Changelog**

:cl:
- tweak: Expanded and rebalanced passive healing across the board
- fix: Made Hydrakin start at their normal body temperature (finally)
- tweak: Adjusted Vox tolerances to being spaced

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Species/base.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/base.yml
@@ -245,15 +245,19 @@
       types:
         Blunt: 0.50 #per second, scales with pressure and other constants.
         Heat: 0.1
-  - type: PassiveDamage # Slight passive regen. Assuming one damage type, comes out to about 4 damage a minute.
+  - type: PassiveDamage # Mono: Slow general passive regen for most types.
     allowedStates:
     - Alive
-    damageCap: 20
+    damageCap: 200
     damage:
       types:
-        Heat: -0.07
+        Heat: -0.05
+        Cold: -0.05
+        Shock: -0.02
+        Radiation: -0.02
+        Poison: -0.02
       groups:
-        Brute: -0.07
+        Brute: -0.15 # Divided among all types whether types exist or not
   - type: Fingerprint
   - type: Blindable
   # Other
--- a/Resources/Prototypes/Entities/Mobs/Species/slime.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/slime.yml
@@ -63,15 +63,21 @@
   - type: Damageable
     damageContainer: Biological
     damageModifierSet: Slime
-  - type: PassiveDamage # Around 8 damage a minute healed
+  - type: PassiveDamage # Mono: More resilient against most damage types and even cellular, very weak to cold.
     allowedStates:
     - Alive
-    damageCap: 65
+    damageCap: 200
     damage:
       types:
-        Heat: -0.14
+        Heat: -0.15
+        Cold: -0.01
+        Shock: -0.05
+        Radiation: -0.05
+        Asphyxiation: -0.05
+        Poison: -0.05
+        Cellular: -0.01
       groups:
-        Brute: -0.14
+        Brute: -0.45 # Divided among all types whether types exist or not
   - type: DamageVisuals
     damageOverlayGroups:
       Brute:
--- a/Resources/Prototypes/Entities/Mobs/Species/vox.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/vox.yml
@@ -33,18 +33,25 @@
   - type: Damageable
     damageContainer: Biological
     damageModifierSet: Vox
-  - type: PassiveDamage
-    # Augment normal health regen to be able to tank some Poison damage
-    # This allows Vox to take their mask off temporarily to eat something without needing a trip to medbay afterwards.
-    allowedStates:
-    - Alive
-    damageCap: 20
+  - type: Barotrauma # Mono: Space-adapted/engineered vox are more resilient against pressure damage
     damage:
       types:
-        Heat: -0.07
-        Poison: -0.2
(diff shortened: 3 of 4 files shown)


UnicornOnLSD:
In discussion with maints.

Redrover1760:
we already have passive healing options, they're not the best but they require work, such as sleeping in a bed (we usually map those on any mildly priced ship) for instance

Tricord isn't too rare, buying med supplies is a pretty common thing, etc.

Passive healing of exotic types is probably not going to fly tho. I don't really think this PR is necessary at all though.

tonotom1:
Shock regen is worth a sniff, it's the only type in current balance where you 100% have to use ointment to be rid of, no bed no passive regen (without platelets)

Maybe it could be added as a bedrest healable damagetype like how cold is

Kiithnaras:
> we already have passive healing options, they're not the best but they require work, such as sleeping in a bed (we usually map those on any mildly priced ship) for instance

That's technically true, but beds present a complication: They require you to do nothing but sit there and not play, no work required. They're great if you want to top up on some of the more annoying damage types to treat while you step away for a bit, but sleeping in a bed for 10-20 minutes because you can't find or make the chems to fix something funky seems a bit of a weak excuse to me.
I think bedhealing is a great supplement to passives, but shouldn't be used to excuse not expanding passives more robustly.
 
> Tricord isn't too rare, buying med supplies is a pretty common thing, etc.

Again, that's true, but one doesn't always have the access to the chems or drugs or have the time to apply them. Keep in mind, these regen rates are at least an order of magnitude less than even the weaker drugs. This change actually reduces the raw rate of Heat regen (but properly buffs Brute types to be equal to to the Heat regen). Even then, it'll take a (human) character 7 minutes and change to recover from _one_ 9mm shot (without armor adjustments) without treatment. I'd hardly consider that unbalanced.

> Passive healing of exotic types is probably not going to fly tho. I don't really think this PR is necessary at all though.

Radiation, Cold, Asphyx, and Shock are anything but exotic. True that healing Asphyx might not be needed since other systems already handle that far better than the rates here. Cellular is the only one here I would consider exotic (and Caustic, but no one fixes Caustic), but I figured it was a neat and niche idea for Slimes exclusively.
Passive regen on Rad makes sure that Uranium ammo doesn't become the meta for eff-you ammo, and also helps the poor sods that don't follow proper safety procedures with DKJr and Pinch gens avoid lengthy and disruptive trips to medical.
Same story with Cold and spacing yourself without your suit sealed, and touching live grilles or cutting wires without insulated gloves (rare, I know, but it happens) with Shock.

Kiithnaras:
Also, bed healing would need a more thorough rework since it has tons of gaps. The (good) medical bed only fixes poison, blunt, and cold - not even heat, slash, or pierce, and definitely not rad or shock or anything else funny.

...that's for sure a project for me to jump on though in my manic insomnia, though...

OneZerooo0:
Mald PR - we have passive healing, it’s called “drink chems”

Kiithnaras:
> Mald PR - we have passive healing, it’s called “drink chems”

Unnecessarily derogatory and not even slightly accurate. I feel like you either didn't study or didn't comprehend the changes and/or just dismissed it based on the title.
Also, as I stated, not everyone has access to those chems, or the specific chems needed, at all times. Even when one does have chems readily on hand, the chems are powerful but short-lasting by comparison, meaning that they are not passive but reactive, and almost always require usage in five-unit increments unless one is willing to go through the excess effort of transferring tiny amounts into vials.

Finally, just to reiterate, these truly passive recovery rates are and are meant to be very slow, less than what even bedhealing can provide, let alone chems. It's meant to mimic what upstream is doing with offmed and passive healing in a more subdued way so that everyone doesn't need to stop playing the game to swing by medical for every minor injury or malady.

Maybe it's overtuned how I have it here, but this is what felt like a reasonable starting point for passive healing. When any kind of combat exceeds the current passive threshold, when most damage sources result in crit within seconds, having the ability to just walk it off after 20-30 minutes feels plenty balanced to me in the many situations that merc and civ players find themselves in.

PR #3520, opened 2026-03-22
Cable Destruction Lag Reduction

tonotom1:
## About the PR

Cables do NOT reach damage threshold to self delete when they get rammed.

We might as well just not let them exist as an item at all upon break because this level of damage makes picking them up to do anything with really unpractical.

Also standardizes cable hp to 200 for all types.

## Why / Balance

The server CHOKED broe

image

## Media

Small beans

## Requirements

## How to test

Smash a cable and be refunded nothing

## Breaking changes

Nah

**Changelog**

:cl:
- tweak: Cables no longer drop an item when destroyed, ever. Hopefully this will reduce ram lag. Cable HP also standardized to 200 for all types.


burrrkeye:
afaik cable hp was buffed to have them not be destroyed before the wall over them got killed first

tonotom1:
> afaik cable hp was buffed to have them not be destroyed before the wall over them got killed first

Ye I retained a buffed value. They used to be 50. Then were buffed to 150/250/300. They're all 200 now.

EckoAurum:
Probably for the best, but I will say, i have survived multiple occasions in a completely destroyed ship by salvaging the bits of wire that flung off like that to repair. Though, probably not worth it for the amount of lag.

Redrover1760:
Ramming doesn't really damage cable at all though, it only damages walls and things with an actual collision fixture.

This pr probably doesn't fix anything.

tonotom1:
> Ramming doesn't really damage cable at all though, it only damages walls and things with an actual collision fixture.
> 
> This pr probably doesn't fix anything.


https://github.com/user-attachments/assets/f837426e-737b-4f67-bd0d-2311b4348abc

Found the culprit and now it does

Side effect of snipped lattice also deleting the underlying cable. Might not be ideal?

Redrover1760 (approved):
lets try it

tonotom1:
I just noticed that a more sophisticated cleanup system for collisions was merged. I was about to close this after discovering such. I think we should try that before going nuclear with my less sophisticated approach.

PR #3532, opened 2026-03-22
Revert "Cable Destruction Lag Reduction"

tonotom1:
Partially Reverts Monolith-Station/Monolith#3520

A more sophisticated system was added so we might not need to go as far as to delete the cable degridding mechanic entirely.

https://github.com/Monolith-Station/Monolith/pull/3415


tonotom1:
Actually we can keep the yml change since it's unrelated

OneZerooo0 (approved):

PR #3557, opened 2026-03-25
Removes M92-X Thermals

Redrover1760:
## About the PR

Removes the thermals on the M92-X tacsuit helmet

## Why / Balance

the tacsuit helmet currently has medhud, sechud, thermals, and NVGs all built in.

This leaves only one reasonable item to slot into eye slot left, sunglasses, to give flash immunity. 

The helmet having 4 huds, alongside both NVGs and thermals at the same time is feature creep. There should be a choice between thermals and sunglasses, here. So this will create that.

If this results in the M92-X being underpowered, that can be addressed, although that's unlikely. PDV hardsuits have very few huds in comparison.

Besides, Thermals is more of a PDV thing, it shouldn't be built into a TSFMC hardsuit, it directly goes against the theming of the faction.

## Requirements

## Breaking changes

**Changelog**

:cl:
- remove: Removes the TSFMC M92-X thermals.

### DIFF
--- a/Resources/Prototypes/_Mono/Entities/Clothing/Head/Hardsuits/tsfmc.yml
+++ b/Resources/Prototypes/_Mono/Entities/Clothing/Head/Hardsuits/tsfmc.yml
@@ -93,10 +93,6 @@
     lowPressureMultiplier: 1000
   - type: EyeProtection
     protectionTime: 15
-  - type: ThermalVision
-    pulseTime: 1.5
-    isEquipment: true
-    toggleAction: PulseThermalVision
   - type: NightVision
     flashDurationMultiplier: 2.75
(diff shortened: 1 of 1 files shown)


Mrrpip:
It won't, you can wear a welding gas mask.

Or if you kill a PDV, loot a watchdog respirator for both the 5% protection and flash protection and eyes free for thermals.

Redrover1760:
> It won't, you can wear a welding gas mask.
> 
> Or if you kill a PDV, loot a watchdog respirator for both the 5% protection and flash protection and eyes free for thermals.

fair point

Redrover1760:
ngl this should not be a single thing and instead be handled as a collective if changes were to be made

PR #3579, opened 2026-03-27
No More Crawling for Non-Targetted Evasion

Ghost581X:
## About the PR
Fixes crawling making you entirely invulnerable to things like shipgun fire.
(ports https://github.com/Sector-Crescent/Hullrot/pull/345 )
## Why / Balance
Apparently it's a bug.

## Media
https://github.com/user-attachments/assets/7b62b1f2-7d46-413f-b0df-0f9468e2861b

## Requirements
- [ ] I have read relevant guidelines/documentation to this PR found on [our devwiki](https://monolith-station.github.io/mono-docs/).
im illiterate help

## How to test
try shooting at a crawling urist vs one that got critted, bullets should pass over those in crit/dead but not over those that're crawling

## Breaking changes
none

:cl:
- fix: Crawling no longer makes you evade shipgun fire.


Redrover1760:
Guns not hitting crawling people is not a bug. Please update your pr with proper justification accordingly

I'd say shipguns are an oversight though.

Shipguns should hit crawling while non shipguns shouldn't without proper aiming

Czuat:
Won't this make it that if you stand on top a corpse and try to shoot someone far away, you will hit the corpse instead? this is ass in that case.

Lokilife:
I don't think this could be a problem. The only problem that I knew is when people were laying down in space in voidboots and this was letting them to avoid shipgun projectiles.
But I updated (#3551) lay logic so it should be impossible to lay down in space.

ThanosDeGraf:
Avoiding bullets by lying down is entirely intentional, that way you and your friend can shoot over one another while in exchange being sitting ducks. So not sure what you mean by "apperantly it's a bug."

tonotom1:
Agree small arms should follow current behavior and shipguns should just delete you regardless of standing position

Redrover1760 (changes requested):
make only shipgun projectiles target people laying down and nothing else if you are able

otherwise, this is not a bug so, close the pr

Ilya246:
we changed our stance on this

PR #3602, opened 2026-03-29
Adc Piloting Boost

Ilya246:
## About the PR
see cl

## Why / Balance
ADC is currently useless so
might also let it slowly autorepair the grid later

**Changelog**

:cl:
- add: ADCs now get piloting boosts (bonus acceleration, turn, max speed) to give them an advantage over ADMs.

### DIFF
--- a/Content.Server/Physics/Controllers/MoverController.cs
+++ b/Content.Server/Physics/Controllers/MoverController.cs
@@ -274,11 +274,19 @@ public sealed class MoverController : SharedMoverController
     //
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


EckoAurum:
the ADC's auto pilot wont get the same buff and will travel slower than it can, but that doesn't really seem like a problem.

Ilya246:
i considered this and i think intentional

Redrover1760 (approved):

NazrinNya:
larp

PR #3608, opened 2026-03-29
[April 1st] Buffs Medipens

mikusssssss:
:cl:
- add: American Supplements


Ilya246 (commented):
feels like this can be done funnier, maybe put meth or something better

mikusssssss:
if its meth then people will use it to move better, fun isnt allowed.

Redrover1760 (changes requested):
meh lame

mikusssssss:
ok

PR #3617, opened 2026-03-29
Allow Scrubbers to Harvest Planetary Atmosphere Tiles

chilla55:
## Summary

This changes fixed air scrubbers so they can harvest gas from planetary map-atmosphere tiles.

Previously, scrubbers rejected immutable gas sources outright, which meant they could not pull from outdoor planetary atmosphere. This patch narrows that restriction so fixed scrubbers can use immutable gas only when the source tile is a non-space map-atmosphere tile.

## What changed

- Added an atmos helper to identify whether a tile is using map atmosphere.
- Updated fixed scrubber processing to:
  - allow immutable-source harvesting on non-space map-atmosphere tiles
  - keep normal immutable-source rejection everywhere else
  - disable WideNet expansion while harvesting map atmosphere so collection stays single-tile
- Left portable scrubber behavior unchanged.

## Behavior

- Fixed scrubbers can now harvest outdoor planetary atmosphere.
- They still cannot harvest arbitrary immutable mixtures.
- WideNet does not fan out across outdoor planetary tiles.
- Portable scrubbers still behave as before.

## Testing

- Built `Content.Server` successfully.
- Verified file diagnostics were clean for the touched files.

## Notes

This is intended as a minimal, targeted reintroduction of atmospheric harvesting using the existing scrubber machine, without adding new art or a separate dedicated harvester.


NazrinNya:
wrong repo brodie

NazrinNya:
also frontier will deny that

chilla55:
sry just noticed

PR #3631, opened 2026-03-29
Remove Repeat Work in ThermalSignatureSystem

Alwayswannahunt:
## About the PR

Well 4 ms is enormous for this type of system. So... Let's clear some obvious things!
I rarely check github, so feel free to push anything into branch

## Why / Balance

Better performance

## Requirements

## How to test

Zero clue...

I spawned shuttle, ftl'ed it, deleted it, spawned again, rocket spam it into dust. (you know, no trust to component events)
Also stopped in debug and happily noticed absence of 1e-7 heat entities update (while some ships heat  ~1e9)

### DIFF
--- a/Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
+++ b/Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
@@ -7,6 +7,7 @@ using Content.Shared.Shuttles.Components;
 using Content.Shared.Shuttles.Systems;
 using Content.Shared.Weapons.Ranged.Components;
 using Content.Shared.Weapons.Ranged.Systems;
+using Robust.Shared.Map.Components;
 using Robust.Shared.Timing;
 
 namespace Content.Server._Mono.Detection;
@@ -23,19 +24,17 @@ public sealed class ThermalSignatureSystem : EntitySystem
     private static readonly TimeSpan UpdateInterval = TimeSpan.FromSeconds(UpdateIntervalSeconds);
     private TimeSpan _nextUpdateTime;
 
-    private const float HeatChangeThreshold = 1.04f;
+    private const float HeatChangeThreshold = 1.02f;
 
     private EntityQuery<ThermalSignatureComponent> _sigQuery;
     private EntityQuery<GunComponent> _gunQuery;
-
-    private Dictionary<EntityUid, ThermalSignatureComponent> _gridCompMap = new();
+    private EntityQuery<MapGridComponent> _mapGridQuery;
 
     public override void Initialize()
     {
         base.Initialize();
 
         SubscribeLocalEvent<GridInitializeEvent>(OnGridInitialized);
-        SubscribeLocalEvent<ThermalSignatureComponent, ComponentShutdown>(OnThermalSignatureShutdown);
 
         // some of this could also be handled in shared but there's no point since PVS is a thing
         SubscribeLocalEvent<MachineThermalSignatureComponent, GetThermalSignatureEvent>(OnMachineGetSignature);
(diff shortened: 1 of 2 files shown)


Ilya246:
the 4ms is fake, grafana misreports systems with an update timer to take as much as they would without a timer

Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
```diff
@@ -33,6 +34,9 @@ public override void Initialize()
     {
         base.Initialize();
 
+        SubscribeLocalEvent<MapGridComponent, MapInitEvent>(OnMapGridMapInit);
```
Ilya246: this should be on comp init

Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
```diff
@@ -87,32 +104,27 @@ private void OnFTLGetSignature(Entity<FTLDriveComponent> ent, ref GetThermalSign
...
         while (query.MoveNext(out var uid, out var sigComp))
         {
-            var ev = new GetThermalSignatureEvent(interval);
+            var ev = new GetThermalSignatureEvent();
             RaiseLocalEvent(uid, ref ev);
+
             sigComp.StoredHeat += ev.Signature * interval;
+
+            // threshold should be really small value so client/server desync wont be a problem
+            if (sigComp.StoredHeat < HeatThreshold)
```
Ilya246: this should track if heat changed instead, have a LastSyncedHeat (explicitly no DataField, VV only) value in the comp and only sync if we're off by 1.04 (check via division) or so

Ilya246 (changes requested):

Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
Alwayswannahunt: This condition is a guard of endlessly multiplying 1e-6 floats, and your proposal ruins it.
Could you please describe what you actually want? Keep in mind that all actually ships has heat around 1e9 and much more, so it never affects anything really valuable

Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
Alwayswannahunt: fix all prototypes any% begins now...

Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
Ilya246: i don't see the issue with multiplying floats, which is basically free? and what i want is for it to not sync over ships which don't change heat

Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
Ilya246: you can ensurecomp on mapinit but being added to list has to be on compinit

Ilya246: also i don't think the list is even needed, just EQE should be cheap enough

Alwayswannahunt: done

Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
Alwayswannahunt: done-ish

Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
```diff
@@ -87,48 +95,46 @@ private void OnFTLGetSignature(Entity<FTLDriveComponent> ent, ref GetThermalSign
...
-        var interval = _updateInterval;
 
-        _gridCompMap.Clear();
+        _nextUpdateTime = _timing.CurTime + UpdateInterval;
 
-        var gridQuery = EntityQueryEnumerator<MapGridComponent>();
-        while (gridQuery.MoveNext(out var uid, out _))
+        var gridQuery = EntityQueryEnumerator<MapGridComponent, ThermalSignatureComponent>();
+        while (gridQuery.MoveNext(out _, out _, out var gridSigComp))
         {
```
Ilya246, line 105: just keep the _sigQuery TryComp and EnsureComp in the loop

Ilya246 (changes requested):

Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
Ilya246: actually nevermind it's fine

Content.Server/_Mono/Detection/ThermalSignatureSystem.cs
```diff
@@ -87,48 +95,46 @@ private void OnFTLGetSignature(Entity<FTLDriveComponent> ent, ref GetThermalSign
...
-        {
-            Dirty(uid, sigComp); // sync to client
+            // threshold should be really small value so client/server desync wont be a problem
+            if (sigComp.TotalHeat <= sigComp.LastUpdateHeat * HeatChangeThreshold
+                && sigComp.TotalHeat >= sigComp.LastUpdateHeat / HeatChangeThreshold)
+                continue;
+
+            sigComp.LastUpdateHeat = sigComp.TotalHeat;
+
+            Dirty(uid, sigComp);
```
Ilya246: should only run for grids since client currently does nothing with non-grid heat

Ilya246 (approved):
good to go if tested

Ilya246:
ok time to test this in april1st publish

PR #3647, opened 2026-03-30
[April 1st] Modify ADS Laws

mikusssssss:
Long requested by ADS players
**Changelog**
:cl:
- fix: Fixed bad ADS laws.


Redrover1760:
I don't think we need ADS randomly murdering POIs, faction bases, safezones and civ ships, even if it is for april fools.

mikusssssss:
> I don't think we need ADS randomly murdering POIs, faction bases, safezones and civ ships, even if it is for april fools.

this is why im now making a pr to make resomi + hydrakin take 65% more damage from everything

PR #3679, opened 2026-04-03
Fix Borg Hand Whitelist Bug

### LINKED ISSUE #2835: Borg can have Hand
## Description
with some module, it is possible to make it bug and let the borg have hand

**Reproduction**
have a module with a beaker, part, circuit, something you can drop on the floor, and that fit in a box (like a tool box)
put an item in the box, put it back in your hand, drop it on the floor
you now have a hand slot that let you use any item
if you change module with the hand slot not fill the module glitch and become broken

EckoAurum:
## About the PR
Closes #2835 

This system to give borgs-psuedo hands should/will be replaced by upstream eventually, but this bug shouldn't exist in the meantime.

I believe what was happening is, this version of borg hands is using ghost items to track swaps, pickups, and putdowns. SwapPlaceholder is getting called too early by prematurely setting Enabled to true, which fucks with containers. This, I believe, ends up putting the ghost item in the bag when attempting to retrieve the whitelisted item, breaking it and causing an invalid metadata flag error.

Now we're doing the swap after the pickup is complete.

## Why / Balance
I think this bug has become fairly prominent even though it's been present for ages.

## Requirements

## How to test

- Control a cyborg in dev environment.
- Select Engineering and select your building module
- Swap to the capacitor and place it into a bag then pull it out again.

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Cyborgs obtaining non-whitelisted hands.

### DIFF
--- a/Content.Shared/_NF/Interaction/Systems/HandPlaceholderSystem.cs
+++ b/Content.Shared/_NF/Interaction/Systems/HandPlaceholderSystem.cs
@@ -137,7 +137,6 @@ public sealed partial class HandPlaceholderSystem : EntitySystem
             return;
 
         SetPlaceholder(target, ent);
-        SetEnabled(target, true);
 
         SetEnabled(ent, false); // allow inserting into the source container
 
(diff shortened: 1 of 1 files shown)


Ilya246 (approved):
if you tested it sure i guess

PR #3702, opened 2026-04-05
Bolt Down Camelot's Military Assembler + Fax + Silo

Ghost581X:
## About the PR
Makes Camelot's fax, material silo and military assembler stealing proof

## Why / Balance
There's only a limited amount of military assemblers and having the only guaranteed one to exist in the round taken by some random guy is not fun.
The fax is USSP's so I think it's fine to also make it unanchorable.
Material silo doesn't matter all that much but I included it anyways since it's for the station.

## Requirements

## How to test
Try stealing Camelot's fax, silo or military assembler.

## Breaking changes
You can't steal them anymore.

**Changelog**
:cl:
- tweak: Camelot secured their military assembler and communication line with extra bolts.


EckoAurum:
This will not work as intended. 
You will have to make new prototypes for these. Just removing the anchorable component in mapping will work for most players, but the astute will simply crowbar them into machine frames, screw it back together, and _then_ steal them.

EckoAurum (changes requested):
Basically same thing, this will not achieved the desired result. A separate prototype is needed. As long as you don't include the construction graph in the new prototype with a different ID it shouldn't be able to be deconstructed. This will involve a bit of juggling with parenting.

NazrinNya (approved):

PR #3723, opened 2026-04-07
Kill Gunnery Server Thermal Sigs

Ilya246:
so if you fly out in a milship it isn't instantly getting hunted down and dying anymore
i am unsure about the full consequences of this but we'll find out

**Changelog**

:cl:
- tweak: Gunnery servers no longer passively make a large thermal signature.

### DIFF
--- a/Resources/Prototypes/_Mono/Entities/Structures/Machines/FireControl/gunnery.yml
+++ b/Resources/Prototypes/_Mono/Entities/Structures/Machines/FireControl/gunnery.yml
@@ -53,9 +53,6 @@
     board: MachineGCSLowCircuitboard
   - type: FireControlServer
     processingPower: 24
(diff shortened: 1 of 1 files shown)


burrrkeye:
cant you just revert them to their original thermal signatures? where a low gunnery kept you hidden up to maximum 412 meters instead of total full stealth

Ilya246:
> cant you just revert them to their original thermal signatures? where a low gunnery kept you hidden up to maximum 412 meters instead of total full stealth

we'll see how this goes first, i kinda want to make gunneries eat power instead but that'll need ship changes

Redrover1760 (approved):
shipguns still create signatures when firing, having more/bigger guns shooting is more heat

thrusters too

power generation as well, same as above

we don't need passive heat gen from gunneries yeah

PR #3737, opened 2026-04-09
Add Compass and Azimuth Ring to Shuttle Radar

ziz11:
## What
Adds two new orientation aids to the shuttle radar UI:
- a small compass for quick orientation
- an outer azimuth ring for precise bearing readout

Also includes visual tuning for readability:
- gray azimuth labels and ticks
- inward-facing ticks only
- no outer segmented ring fragments

## Why
The shuttle radar previously had no dedicated compass or azimuth bearing scale, which made orientation and precise bearing readout harder during navigation.

This change separates:
- quick orientation via the compass
- precise bearing readout via the azimuth ring

## Notes
- In No Rotate, the azimuth ring is north-up
- In Rotate, the radar follows the selected rotation behavior
- The ring is intentionally subtle so it does not compete with contacts

### DIFF
--- a/Content.Client/Shuttles/UI/BaseShuttleControl.xaml.cs
+++ b/Content.Client/Shuttles/UI/BaseShuttleControl.xaml.cs
@@ -138,6 +138,216 @@ public partial class BaseShuttleControl : MapGridControl
     }
     // End Frontier Corvax
 
+    // Mono
+    protected void DrawAzimuthScale(DrawingHandleScreen handle, Angle rotation)
+    {
+        const int azimuthMinorTickStepDegrees = 10;
+        const int azimuthMajorTickInterval = 3;
+        const int azimuthCardinalTickInterval = 9;
+        const float edgeInset = 10f;
+        const float tickInset = 4f;
+        const float minorTickLength = 5f;
+        const float majorTickLength = 9f;
+        const float cardinalTickLength = 12f;
+        const float labelPadding = 3f;
+        const float labelOffset = tickInset + cardinalTickLength + labelPadding;
+        const float regularLabelScale = 0.62f;
+        const float cardinalLabelScale = 0.72f;
+
+        var origin = MidPointVector;
+        var viewportSize = new Vector2(PixelWidth, PixelHeight);
+
+        var subtleGray = Color.FromHex("#676767");
+        var northAccent = Color.FromHex("#7F7442");
+        var eastAccent = Color.FromHex("#456273");
+        var baseTickColor = subtleGray.WithAlpha(0.34f);
+        var majorTickColor = subtleGray.WithAlpha(0.5f);
+        var cardinalTickColor = subtleGray.WithAlpha(0.68f);
+        var regularLabelColor = subtleGray.WithAlpha(0.76f);
+        var cardinalLabelColor = subtleGray.WithAlpha(0.9f);
+        var labelShadowColor = Color.Black.WithAlpha(0.36f);
+
+        for (var step = 0; step < 360 / azimuthMinorTickStepDegrees; step++)
+        {
+            var degrees = step * azimuthMinorTickStepDegrees;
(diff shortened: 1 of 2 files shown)


OneZerooo0 (commented):
holy mother of based

Content.Client/Shuttles/UI/BaseShuttleControl.xaml.cs
```diff
@@ -136,6 +145,110 @@ protected void DrawNorthLine(DrawingHandleScreen handle, Angle angle)
...
+        handle.DrawLine(end, end - direction * headLength - left * (headLength * 0.45f), color);
+    }
+
+    private static Vector2 GetAzimuthDirection(Angle baseAngle, float azimuthDegrees)
+    {
+        var angle = baseAngle + Angle.FromDegrees(azimuthDegrees);
+        var radians = (float) angle.Theta - MathF.PI / 2f;
+        return new Vector2(MathF.Cos(radians), MathF.Sin(radians));
+    }
     // End Frontier Corvax
```
OneZerooo0: Is this change from frontier corvax? If not, make sure you properly seperate _Mono namespace

OneZerooo0 (changes requested):

Content.Client/Shuttles/UI/BaseShuttleControl.xaml.cs
```diff
@@ -24,6 +24,15 @@ namespace Content.Client.Shuttles.UI;
...
 {
+    private const int AzimuthMinorTickStepDegrees = 10;
+    private const int AzimuthMajorTickInterval = 3;
+    private const int AzimuthCardinalTickInterval = 9;
+    private static readonly string[] AzimuthLabels =
+    [
+        "000", "030", "060", "090", "120", "150",
+        "180", "210", "240", "270", "300", "330"
+    ];
+
```
Ilya246: place those lower for code style consistency
also can place them in-method since not used anywhere else

Content.Client/Shuttles/UI/BaseShuttleControl.xaml.cs
ziz11: - Not copied from Frontier Corvax. FrontierComments Were in code before me
- That functions were developed from scratch.
- Shared drawing code stays in BaseShuttleControl, while _Mono-specific behavior remains in FireControlNavControl

Content.Client/Shuttles/UI/BaseShuttleControl.xaml.cs
ziz11: Moved the azimuth-only constants into DrawAzimuthScale since they are only used there.

Content.Client/Shuttles/UI/BaseShuttleControl.xaml.cs
ziz11: I marked the inserted blocks as Mono to keep the separation explicit for merges.

Content.Client/_Mono/FireControl/UI/FireControlNavControl.cs
```diff
@@ -46,6 +46,12 @@ public FireControlNavControl() : base(64f, 512f, 512f)
         _transform = EntManager.System<SharedTransformSystem>();
     }
 
+    // Mono
+    protected override bool ShouldShowCompassOverlay(RadarConsoleComponent? radar)
+    {
+        return false;
+    }
+
```
Ilya246: i'm thinking maybe this could instead just check `_rotation == Angle.Zero` so it works for all consoles that don't rotate, which includes radar consoles

ziz11: Done, removed the gunnery-specific override.

Azimuth now draws from the shared radar path, including worldRot == Angle.Zero. Compass is hidden for north-locked views like gunnery.

Ilya246 (approved):
seems fine

PR #3739, opened 2026-04-09
Buffs Ore Rates

Redrover1760:
## About the PR

approximately doubles ore drops from mining.

## Why / Balance

We're still using old frontier values that really don't give all that much per asteroid overall.

Its not really all that worth it to mine asteroids with these rates with how the economy is so tweaked.

Notable: Yes, ship mining and stuff exists and hyper optimizing eating up asteroid after asteroid, but this should make non-ship mining and the like viable too without having to non-stop mine.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Asteroid ore rates doubled.

### DIFF
--- a/Resources/Prototypes/_NF/ore.yml
+++ b/Resources/Prototypes/_NF/ore.yml
@@ -6,64 +6,64 @@
 - type: ore
   id: NFOreSteel
   oreEntity: SteelOre5
-  minOreYield: 1
-  maxOreYield: 5
+  minOreYield: 2
+  maxOreYield: 10
 
 - type: ore
   id: NFOreSpaceQuartz
   oreEntity: SpaceQuartz5
-  minOreYield: 1
-  maxOreYield: 5
+  minOreYield: 2
+  maxOreYield: 10
 
 - type: ore
   id: NFOreCoal
   oreEntity: Coal5
-  minOreYield: 1
-  maxOreYield: 5
+  minOreYield: 2
+  maxOreYield: 10
 
 # Medium yields
 - type: ore
   id: NFOreGold
   oreEntity: GoldOre5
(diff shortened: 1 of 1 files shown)


Ilya246:
i don't think this is needed since mining already yields a lot, we can however increase its money return potentially

ThanosDeGraf:
> i don't think this is needed since mining already yields a lot, we can however increase its money return potentially

You barely increase your profit from ores when processing them into materials. Wouldn't this just make it more optimal to sell buckets of ore without processing them at all? Unless this also increases the value of every material in turn, which would also just buff every other way of obtaining materials.

Ilya246:
yes i meant both ore and materials and that's fine

ThanosDeGraf:
> yes i meant both ore and materials and that's fine

One only buffs ore miner income, the other is a major inflation of the entire economy.

Tryded:
> One only buffs ore miner income, the other is a major inflation of the entire economy.

This PR might buff mining income substantially if they are solely selling to Non-Players, however it actually harms the player materials market by deflating ore prices significantly. Since there is increased supply by basically 100%, other players and major factions will have even less of a need to buy materials from miners as mining by themselves can easily cover all of their needs due to the sudden increase in mining productivity.

Ideally player to player markets should be encouraged as it makes the economy more engaging and important for factions.

Redrover1760:
> yes i meant both ore and materials and that's fine

adjusting material worth increases the cost of everything

including ships

If this was desired, it would probably be best to increase the material cost of exotic stuff (diamonds, bluespace, plasma, etc)

NazrinNya:
nah
just make cost of exotic mats higher idk

PR #3775, opened 2026-04-12
Adds Missing Locale Strings

ondrej008:
so that useless warnings don't get spammed, someone else reported this before idk if they made a github issue


ondrej008:
YOU MUST ASK TO BE PAID AFTER MAINTAINERS CONFIRM IT'S A REAL BUG & YOU'RE THE FIRST TO FIND IT.

ondrej008:
[/Locale/en-US/shuttles/console.ftl]: For id:shuttle-console-undock-all already exist entry of type: Message
[/Locale/en-US/shuttles/console.ftl]: For id:shuttle-console-ftl-lock already exist entry of type: Message

all the locales (except the network tooltip) were already in but the geniuses at Frontier decided to just replace them with hardcoded text that was marked as a locale string which was equivalent to the locale translation... I don't even know at this point what they were doing

NazrinNya (approved):

PR #3781, opened 2026-04-13
Auto Disable Jetpack when Entering Grid with Enabled Magboots

doublechest0:
## About the PR
Auto disable jetpack when entering grid with enabled magboots

## Why / Balance
You can't move when you enter the grid with magboots and jetpack enabled.
## Media
Before:

https://github.com/user-attachments/assets/3de63f6c-80b2-46e6-bc13-306cb486db3f

After

https://github.com/user-attachments/assets/7eb172fa-2e69-40a4-979e-b5a97c249a02

## Requirements

## Breaking changes

**Changelog**

:cl: doublechest
- fix: Fixed not being able to move when entering the grid with jetpacks and magboots enabled


MantasDab360:
[Insert Absolute Cinema]

Redrover1760:
update your main branch to latest and run git submodule update --recursive to update to .NET 10.0 and latest engine version we use

Content.Shared/Clothing/MagbootsSystem.cs
```diff
@@ -86,5 +91,20 @@ private void OnIsWeightless(Entity<MagbootsComponent> ent, ref IsWeightlessEvent
     private void OnIsWeightless(Entity<MagbootsComponent> ent, ref InventoryRelayedEvent<IsWeightlessEvent> args)
     {
         OnIsWeightless(ent, ref args.Args);
+
+        // Mono: Disable jetpack if enabled magboots are on and the user is on a gravity-supporting grid or map. This is done here because the jetpack user component is on the mob, not the magboots.
+
+        if (!_toggle.IsActivated(ent.Owner))
```
Redrover1760: I'm not an expert at C# but

A: Should we be checking to see if args for the event is already handled here?

B: This code could probably be moved into its own method, "HandleActiveJetpack" or so.

Ilya246 (changes requested):
this should instead be in jetpack system, preferably in a subscription to parent change event

doublechest0:
`TryComp<MovedByPressureComponent>(uid, out var movedPressure) && !movedPressure.Enabled`
A way to detect magboots on a jetpack user that I found in other code

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -167,6 +169,12 @@ private void OnJetpackToggle(EntityUid uid, JetpackComponent component, ToggleJe
 
             return;
         }
+        else if (Transform(uid).GridUid != null && TryComp<MovedByPressureComponent>(uid, out var movedPressure) && !movedPressure.Enabled) // Mono
+        {
+            _popup.PopupClient(Loc.GetString("jetpack-no-enabled-magboots"), uid, args.Performer);
+
+            return;
+        }
```
Ilya246: just integrate this into the above `if`

Ilya246 (changes requested):
also use SharedGravitySystem's IsWeightless instead, don't need to check for grid either with it
this will also fix it for voidboots

doublechest0:
I added MagbootsToggledEvent to fix situations where magboots switch while the player is flying on the grid using jetpack

Content.Shared/Clothing/MagbootsSystem.cs
```diff
@@ -88,3 +92,6 @@ private void OnIsWeightless(Entity<MagbootsComponent> ent, ref InventoryRelayedE
         OnIsWeightless(ent, ref args.Args);
     }
 }
+
+[ByRefEvent]
+public record struct MagbootsToggledEvent(EntityUid User, EntityUid Magboots, bool State); // Mono
```
Ilya246: you don't need to pass User it's the entity you raise the event on

Content.Shared/Movement/Systems/SharedJetpackSystem.cs
```diff
@@ -245,6 +250,17 @@ protected virtual bool UserNotParented(EntityUid? user, JetpackComponent compone
             || xform.ParentUid == xform.MapUid;
     }
     // End EE
+
+    // Mono
+    private void OnJetpackUserMagbootsToggled(EntityUid uid, JetpackUserComponent component, ref MagbootsToggledEvent args)
+    {
+        if (!args.State || !IsEnabled(component.Jetpack) || Transform(uid).GridUid == null || !TryComp<JetpackComponent>(component.Jetpack, out var jetpack))
```
Ilya246: this can also just only check if we became non-weightless using IsWeightless

Ilya246 (changes requested):

Ilya246 (approved):
good to go if tested

PR #3788, opened 2026-04-14
Revert "Fix Projectile Raycast"

chuga-git:
This change unintentionally made explosion based shipguns do ~8x less damage. Temporary revert until a fix for the unintended behavior is made.

**Changelog**

:cl:
- fix: Reverts projectile explosion fix, pending a proper patch for off-grid explosions to make them not extremely inconsistent.


Ilya246:
get fluffen rebalance out instead

Redrover1760 (approved):
The way off-grid explosions interact with on-grid tiles makes fixing explosions more complicated than just buffing them from testing.

This has the approval to keep the game playable until a proper fix for off-grid in explosions is made.

PR #3810, opened 2026-04-17
[Port] Mapping Editor - Early Merge 'N Tweaks

lexaSvarshik:
LICENSE: MIT

## About the PR

Ports https://github.com/Reserve-Station/Reserve-Station/pull/82 which is early merge of stale wizden PR, see it for more details.

Mapping simpler now

use `mapping <id> <path>`
or `mappingclientsidesetup` to toggle the gui

`mappingclientsidesetupdisable` to switch back to normal state
## Media

Quick look:
изображение
Mapping templates, disabled rendering of walls, airlocks, markers

изображение
Tiles and decals
## Requirements

## How to test

run `mapping` command, load any map, try messing with menus
## Breaking changes

**Changelog**


Ilya246:
needs to handle our shuttle stuff such as radar edge markers and non-square tiles well

lexaSvarshik:
> needs to handle our shuttle stuff such as radar edge markers and non-square tiles well

will do

Redrover1760:
the original is MIT, so this should be licensed MIT as well (I just put it in the PR it works)

lexaSvarshik:
good

Content.Server/Mapping/MappingManager.cs
```diff
@@ -76,4 +93,47 @@ private void OnMappingSaveMap(MappingSaveMapMessage message)
...
+
+        var path = new ResPath(FavoritesPath);
+        using var writer = _resourceMan.UserData.OpenWriteText(path);
+        var stream = new YamlStream { new(mapping.ToYaml()) };
+        stream.Save(new YamlMappingFix(new Emitter(writer)), false);
+    }
+
+    private void OnMappingFavoritesLoad(MappingFavoritesLoadMessage message)
+    {
+        var path = new ResPath(FavoritesPath);
```
Ilya246, line 128: those should either A) be gatekept by a cvar that enables/disables them and is false by default but true in devenv (there's a devenv config file) B) check admin perms on the sender
currently those could be used on live server to take up disk space

Resources/Locale/ru-RU/_wizden/mapping/mapping-visibility.ftl
Ilya246, line 1: i don't think russian locale is needed?

Ilya246 (changes requested):

Corrupt142 (changes requested):
i didnt look at the code but!

### Mandatory:

Image
- The mirrored arrow has a ERROR sign (probably a sprite error)

Image

- All radar marking showcase sprites are the same, it can be confusing

### Suggestions:

- Visualiser of multitool connections (what machine/door is connected to other machines/doors/buttons, ect)

- add a mass scanner ability button to check the ship out on mass overview

image

- Try to make the text not get cut off so instead of it just dissapearing into the ather it would do this instead: 

Airlock 
Exterior,Locked

image

- The middle mouse rotate should pick the north,south,east,west version of the decals to prevent the missalignment and other visual inconsistencies

UnicornOnLSD (changes requested):
Tested it myself, I'll only mentions issues I have for the sake of efficiency:

- The color hexer is broken, I cannot input the hexcode of a color I want (for example #0055CC) when I tried to paint pipes. It just breaks and autofills another color.

- When searching for specific tiles, the scrolling enus don't show up which is pain since it'll likely be how mappers will be looking for the entire menu and now just all the items spread out (see image) 

Image

Instead of

Image

- While you CAN favorite any kind of tiles, you cannot have one specific item bound to a key. **You need to be able to bind some tiles to action keys like before**

- As corrupt said, a mass scanner view.

- A way to toggle the UI back to regular client on and off especially for testing purposes would be deeply appreciated.

- Possibly showing the entity ID when hovering over them instead of the same thing, that way you can search for similar items easier

Image

- It's a bit weird the old "copy button" is replaced by a new bind which does the same thing for the "pick entity or tile"



It's some of the issues I've seen while testing, however, it's a massive improvement overall.

lexaSvarshik:
изображение
tile search fixed kinda

lexaSvarshik:
tbd mass scanner, exit button, multitool connections and id instead of names when hovering

lexaSvarshik:
mass scanner button is not working and i have zero clue why

Ilya246:
say when ready for rereview

lexaSvarshik:
grr upstream

Ilya246:
upstream decided to send a missile into your PR

lexaSvarshik:
not related testfail, mewge

Ilya246 (commented):
remove `readonly` from any [Dependency] fields and add `partial` to any classes that contain any [Dependency] fields

Ilya246:
- mass scanner button does indeed not work, to make it work just add a new message that client sends to server, and server will open mass scanner on client's entity if client's entity is a mass scanner
- i think the category names should also open the menu on click, not just the arrows
- this starts very zoomed out for me
- should probably automatically switch off lighting on startup
- could automatically place a single steel tile at 0, 0 on startup
- perhaps add a command to simply toggle mapping menu without any side effects, so you can switch to normal aghost while mapping or to mapping mode while not mapping

also
```
[ERRO] net: [::1]:57293: exception in message handler for MappingFavoritesSaveMessage:
System.IO.IOException: The process cannot access the file '/home/ilya/sources/space-station-14/bin/Content.Server/data/mapping_editor_favorites.yml' because it is being used by another process.
```

lexaSvarshik:
>perhaps add a command to simply toggle mapping menu without any side effects, so you can switch to normal aghost while mapping or to mapping mode while not mapping

`scene` can be used for that i believe

>this starts very zoomed out for me
should probably automatically switch off lighting on startup

will fix

>mass scanner button does indeed not work, to make it work just add a new message that client sends to server, and server will open mass scanner on client's entity if client's entity is a mass scanner

will try

PR #3811, opened 2026-04-17
Even More Ship HTN Upgrades

Ilya246:
**Media**
turbobaeg VS minefield

https://github.com/user-attachments/assets/36fb8fee-723d-4c3d-b8ab-56f9e851f19b

**Changelog**

:cl:
- tweak: Significantly upgraded ship AI (autopilot) behavior. Should be less awkward and no longer make easily avoidable mistakes (will still make unavoidable mistakes, though, so don't go going 300 on a Serkesh).


Ilya246:
i don't think anyone else here can review this so

PR #3828, opened 2026-04-19
Fix Projectile Penetration

Ilya246:
## About the PR
see title
also improves how raycast works
also adds capability for projectiles to detonate on final hit rather than on first hit

**Changelog**

:cl:
- fix: Fixed projectile penetration. This means Charon can penetrate weak walls now.


OneZerooo0 (approved):
Based

kasature90 (approved):

PR #3839, opened 2026-04-19
The Great Artifact Inflation

ThanosDeGraf:
## About the PR

Increased artifact node value based on depth. Also changed the base BiasDirection of artifact analyzers from Up to Down.

For example, a fully researched 12 node artifact (which in a dev environment took me 15 minutes of debug-activating it and hoping to branch to the correct node) would result in the artifact being worth 69241 spesos. Yea I threw up in my mouth a little just saying that.
image

## Why / Balance
The base BiasDirection change was something that bothered me since forever. When a new artifact analyzer is spawned or constructed, the bias direction pre-selected would always be Up. The only reason you would want to go up is to re-tread a node line to go for another chain, while Down is what you want to go for the other 90% of the time. This just removes one unnecessary for every artifact analyzer in every science ship you order.

Some friends I brought over to monolith had mentioned that they stopped doing any science here, because every other profession they've tried is much more reliable or more profitable than the thing they like to actually do. I then checked how the speso value of artifacts is calculated and it's abysmal.

In general, first you take all discovered nodes of the artifact. Each node is worth 6500RP at base. Multiply that by a PointsPerNode value, which is also raised by the node's PointDangerMultiplier (node's depth and danger of each the effect and trigger). Then if it's an untriggered node, reduce the node's value by a quarter. Finally, the RP value of every node is totaled up and multiplied by a PriceMultiplier value to get the number you get from an Appraisal Tool.

The issue lies in the PointDangerMultiplier and PriceMultiplier. You can krank either of these values for an increase in price, but I went for PointDangerMultiplier, since it scales really nicely with the complexity of artifacts. Untouched artifacts remain at 650 spesos, simple fully researched ones ~1.5x in price, while fully researched 12 node artifacts ~double. [Edited since PointDangerMultiplier]

## Media

Doesn't make sense to add more because every tiny change makes screenshots useless.
image

## Requirements

## How to test

Enter a local game
Spawn an simple artifact, unlock all nodes.
Appraise the artifact
Open the ArtifactComponent and play with the PointDangerMultiplier value (was originally 1.35), press enter!
Appraise again.

Repeat for a complex artifact.

**Changelog**
:cl: Tuerk
- tweak: Artifact analyzers will now have their bias set to Down by default.
- tweak: Artifacts are worth more research points and credits now. The deeper the node, the more science and credits.
- fix: Crash when artifact analyzers had a lot of biasChance affecting bonuses.


ThanosDeGraf:
Forgot to mention but this also fixes a bug where the server crashes if you use an artifact analyzer enhanced with multiple bluespace manipulators

Content.Server/Xenoarchaeology/XenoArtifacts/ArtifactSystem.cs
```diff
@@ -270,7 +254,7 @@ public void ForceActivateArtifact(EntityUid uid, EntityUid? user = null, Artifac
 
         if (TryComp<BiasedArtifactComponent>(uid, out var bias) &&
             TryComp<TraversalDistorterComponent>(bias.Provider, out var trav) &&
-            _random.Prob(trav.BiasChance) &&
+            _random.Prob(MathF.Min(1.0f,trav.BiasChance)) &&
             this.IsPowered(bias.Provider, EntityManager))
```
NazrinNya, line 258: what exactly it does
also comments

Content.Server/Xenoarchaeology/XenoArtifacts/ArtifactSystem.cs
```diff
@@ -45,12 +45,6 @@ public override void Initialize()
     /// Calculates the price of an artifact based on
     /// how many nodes have been unlocked/triggered
     /// </summary>
-    /// <remarks>
-    /// General balancing (for fully unlocked artifacts):
-    /// Simple (1-2 Nodes): 1-2K
-    /// Medium (5-8 Nodes): 6-7K
-    /// Complex (7-12 Nodes): 10-11K
-    /// </remarks>
```
NazrinNya, line 53: why

Content.Server/Xenoarchaeology/XenoArtifacts/ArtifactSystem.cs
```diff
@@ -59,16 +53,6 @@ private void GetPrice(EntityUid uid, ArtifactComponent component, ref PriceCalcu
...
-    /// <remarks>
-    /// General balancing (for fully unlocked artifacts):
-    /// Simple (1-2 Nodes): ~10K
-    /// Medium (5-8 Nodes): ~30-40K
-    /// Complex (7-12 Nodes): ~60-80K
-    ///
-    /// Simple artifacts should be enough to unlock a few techs.
-    /// Medium should get you partway through a tree.
-    /// Complex should get you through a full tree and then some.
-    /// </remarks>
```
NazrinNya, line 71: why

NazrinNya (changes requested):

NazrinNya:
Also i would probably make artifacts not give that much spesos but give good amount of RP too

MantasDab360:
this would go hard with the new xenoarch but yeah, we love money.

Content.Server/Xenoarchaeology/XenoArtifacts/ArtifactSystem.cs
ThanosDeGraf: Monolith has adjusted the conversion of reseaech points -> speso value to like 3.2x, so those remarks didn't make sense anymore. I could sit there for an hour or two and hand-calculate how much a fully researched artifact of each size pays out, or just remove the outdated remark as I did here.

Content.Server/Xenoarchaeology/XenoArtifacts/ArtifactSystem.cs
ThanosDeGraf: See my previous remark, but in this case the mono buffed research output of artifacts.

Content.Server/Xenoarchaeology/XenoArtifacts/ArtifactSystem.cs
ThanosDeGraf: You are right it's missing a mono comment.

Therr was a crash where if you upgrade the artifact analyzer with better manipulators, it would increase the BiasChance to above 1.0 (~1.12 with three bluespace manips). _random.Prob throws an exception when the argument is outside of 0.0 - 1.0, then crash the server.

With the added .Min() it clamps the BiasChance to 1.0 so the crash doesn't happen.

ThanosDeGraf:
I'll be home in about an hour, so I'll add the missing comments and corrected changelog then. :3

> Also i would probably make artifacts not give that much spesos but give good amount of RP too

With my changes, more "dangerous" nodes will pay out much much more RP than before. Danger in this case is just the average of the node's depth, its effect targetDepth and trigger targetDepth.

Since the RP output is buffed, I can go back and reduce the conversion from 0.4x to smth like 0.3x or 0.35x.

## About the "effort" -> RP/Speso conversion:

I had to sit in the dev environment for fifteen minutes, using the debug activate-artifact verb to speedrun a max-sized 12 node artifact, plus I upgraded the artifact analyzer with a bluespace capacitor so the scans took 2 seconds instead of 10. After I was done, the un-buffed speso value came to about 70k. The artifact I had also had an early EMP node, which slowed me down even more. I wanted to weep so bad when I saw tze appraisal.

Content.Server/Xenoarchaeology/XenoArtifacts/ArtifactComponent.cs
```diff
@@ -58,7 +58,7 @@ public sealed partial class ArtifactComponent : Component
     /// to determine the monetary value of the artifact
     /// </summary>
     [DataField("priceMultiplier"), ViewVariables(VVAccess.ReadWrite)]
-    public float PriceMultiplier = 0.4f; // Frontier: 0.175 < 0.4
+    public float PriceMultiplier = 0.4f; // Frontier: 0.175 < 0.35 //Mono: 0.4f -> 0.35f (PointDangerMultiplier increased so we can reduce this a bit)
```
Redrover1760: these comments no longer match the numbers

Content.Server/Xenoarchaeology/Equipment/Components/TraversalDistorterComponent.cs
```diff
@@ -23,7 +23,7 @@ public sealed partial class TraversalDistorterComponent : Component
     public float PartRatingBiasChance = 1.1f;
 
     [ViewVariables(VVAccess.ReadWrite)]
-    public BiasDirection BiasDirection = BiasDirection.Up;
+    public BiasDirection BiasDirection = BiasDirection.Down; // Mono
```
Redrover1760: why

Also, if you make changes like these to code, please state them in the PR body itself please and ideally log why you did them.

Redrover1760 (changes requested):

Content.Server/Xenoarchaeology/XenoArtifacts/ArtifactComponent.cs
ThanosDeGraf: Okay that's odd, I thought I *did* change em to 0.35. Maybe forgot to commit it.

Content.Server/Xenoarchaeology/Equipment/Components/TraversalDistorterComponent.cs
ThanosDeGraf: ~~This is the most minor QoL change that 99% of players will not see because they have not touched an artifact in their life~~ Yea I can do that.

Redrover1760:
the main issue is that changes should be documented

if I had to guess it just defaults it to down instead of up

ThanosDeGraf:
> the main issue is that changes should be documented
> 
> if I had to guess it just defaults it to down instead of up

In my future PRs I'll try to document my changes better. Sorry about that! ~~bites on my own tongue to not bring up dozens of worse PRs~~

ThanosDeGraf:
Media to show what I claim the BaseBias change does actually does what I hope

https://github.com/user-attachments/assets/039c09f1-facf-4f58-9e64-919c87444633

You can see how the spawned in and un-flatpacked artifact analyzer come in with BiasDirection set to Down already. Meanwhile on master it would be Up.

Content.Server/Xenoarchaeology/XenoArtifacts/ArtifactComponent.cs
```diff
@@ -77,7 +77,7 @@ public sealed partial class ArtifactComponent : Component
     /// Used for calculating the research point value of an artifact node.
     /// </summary>
     [DataField("pointDangerMultiplier"), ViewVariables(VVAccess.ReadWrite)]
-    public float PointDangerMultiplier = 1.35f;
+    public float PointDangerMultiplier = 2.25f; // Mono: 1.35 < 2.25 //Note: Why tf does frontier use < instead of -> ???
```
NazrinNya: I think it should be 1.6-1.8 instead, since growth becomes exponential basically.
also increase price multi then

ThanosDeGraf: > [...] since growth becomes exponential basically.

I guess so. I'm happy with putting the price mult back up to 0.4 because then untouched arti's go back to a clean 650 spesos.

NazrinNya (approved):
Ok

ThanosDeGraf:
Updated changelog

ThanosDeGraf:
No test fails, let's GET IT

The-Aisen:
Commenting as a science player, please don't let this PR fall into purgatory; artifacts need a little love, and by itself the science gameplay loop doesn't get many rewards to the scientist (besides the occasional omni/cognizine artifact); a little extra money would help getting science into a nicer spot compared to more lucrative gameloops

PR #3844, opened 2026-04-20
Improves Sloppy Code

Redrover1760:
## Requirements

## Breaking changes

**Changelog**


Content.Shared/Follower/FollowerSystem.cs
```diff
@@ -173,6 +173,9 @@ private void OnFollowedPolymorphed(Entity<FollowedComponent> entity, ref Polymor
     /// <param name="entity">The entity to be followed</param>
     public void StartFollowingEntity(EntityUid follower, EntityUid entity)
     {
+        if (TerminatingOrDeleted(follower) || TerminatingOrDeleted(entity)) // Mono: Ensure the entities exist.
```
Ilya246: why is this needed? this seems like there's an underlying issue that causes deleted entities to be fed to this and which should be fixed instead

Redrover1760: Probably. I'll take a look.

And its cause of some errors I found

PR #3850, opened 2026-04-21
[Faction Rework] TSFMC Meridian 2

OneZerooo0:
## About the PR

Adds the TSFMC battlecarrier Meridian, replacing the halcyon in-sector
Halcyon will be moved to being an outer-sector base.

WIP - Needs more accompanying content (LPC update, gateway system, two versions for two different game modes, move halcyon)

Primarily up for map review
## Why / Balance

Faction rework content
Halcyon is beautiful but also outdated
## Media

tsfmcmeridian-0
## Requirements

## How to test

Spawn in and see
## Breaking changes

**Changelog**

Not player facing yet


kasature90:
i feel like this is too small for a base, even a secondary one
betting it will just be used for ram tbh

kasature90:
(jupiter 2)

burrrkeye:
> i feel like this is too small for a base, even a secondary one betting it will just be used for ram tbh

incorrect, this is their ONLY base, so it will not, unless they choose to do so and  throw their only base away

OneZerooo0:
It’s going to be the only really fast way to get into the sector even with a full outer sector base - if people just use it for ramslop I’ll make changes

waphun:
Why does it look like that? The general shape makes it look unintentional, like it was cobbled together from different ships but it's clear that you've made conscious decision to make it look like that (those southern docks are the most telling). It certainly doesn't look like the flagship of the TSFMC's operations in the sector.
If this is meant to be fast then you shouldn't just think about forward speed but also turning, strafe and reverse where the Meridian struggles a lot.
It is comically easy to hijack, it took me less than 2 minutes to do with an authentication disruptor (20 DC or research) and two pieces of C4 (10 DC or research). 
takeover
(The red Xs are where I used an auth disruptor and orange Xs are where I used C4, the branching paths are alternative routes)
The Dymeres' have terrible LoS and there are no points where they converge which is bad because they are the only long-range weaponry on this. If the PDV show up with Hydras, Cyrexas or even Leviathans they can easily take firing positions where the meridian won't be able to fight back (The Saintie is especially deadly because it can easily outrun/outstrafe the meridian while staying in its blind spots). If you are only going to use two Dymeres then make sure that they can at least fire backwards (you could do that by placing them at the tip of the wings rather than on the sides).
Dymere-range
As the others have pointed out, this is way too small to be the TSFMC base, there just aren't enough ports for people to dock their ships even on midpop and the ports that are there too close to each other making docking very precarious. 
Accessible-Parking
(I realised during writing this that I should have spawned in an Andromeda because one of those always shows up during a round and they are horrendous to park)
Also, update your media. There are discrepancies between the current screenshots/renders and what's in-game (the highlights I made were based on the in-game version).
Overall I could have nitpicked minor violations of turrets placements or over-exposed platings but I think there are more fundamental issues with this that warrant at least a partial redesign. Make it bigger, make it look like it's a mighty, grandiose battle station and work your way backwards from the shape to the internals.

kasature90:
i agree w/ the review above, + for a battlecarrier flagship this seems kinda mediocre at carrier-ing and bad at battling?

OneZerooo0:
Keep in mind that this is more of a secondary outpost - though I will expand/up-armor it

projectkepler-ru:
this ship looks fucking ugly

Snuke1942:
How long to wait?

projectkepler-ru:
> How long to wait?

soon. I will be helping zero with it

EldritchPineapple:
just a week away

PR #3853, opened 2026-04-22
Generation 0 Memory Optimizations

Redrover1760:
## About the PR

optimizes gen0 memory by reusing memory instead of allocating.

Found through dotMemory full allocations analysis

Archive gasmixture: If air gasmixture archive already exists, simply replace it instead of creating a new GasMixture.

Pathfinding: Comment out unused stopwatching, don't make a hashset for tile neighbors unless necessary (i.e., if they exist, AddNeighbors adds them.) 

PathfindingData, don't compare enums with Equals(). Slower and boxes an enum in order to confirm both enums are of the same type, creating unnecessary heap. Use == instead.

## Why / Balance

optimizations

## Media

before:
Heres another

after:
got it down

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Atmos/NPC short term memory optimizations


kasature90 (approved):
goodbye neighbor

PR #3870, opened 2026-04-24
Fix ASM-220s From Blowing Themselves Up

kasature90:
## About the PR
lowers ASM-220 HE/ECM collision radius to be the same as LOSAT (0.3)

## Why / Balance
missile launchers adjacent to eachother (and other configurations but this is the most notable i think) would just blow up the launcher if you fired them, hitbox was way too big

## Requirements

## Breaking changes

**Changelog**
:cl:
- fix: ASM-220 torpedoes should be less likely to blow themselves or their own launcher up now.


burrrkeye:
imo this should be saved by fixing mapping issues and not reducing hitbox

OneZerooo0 (approved):

PR #3873, opened 2026-04-24
Sale Price Test and Material Upprice

Ilya246:
## About the PR
- adds test to see if any recipe does not multiply the sale price of its inputs by more than 4x
- upprices every material because they were laughably cheap

**Changelog**

:cl:
- tweak: Materials and ore sell for significantly more now.

### DIFF
--- a/Content.IntegrationTests/Tests/_Mono/ManufacturePriceTest.cs
+++ b/Content.IntegrationTests/Tests/_Mono/ManufacturePriceTest.cs
@@ -1,0 +1,88 @@
+using Content.Server.Cargo.Systems;
+using Content.Shared.Lathe;
+using Content.Shared.Lathe.Prototypes;
+using Content.Shared.Materials;
+using Content.Shared.Research.Prototypes;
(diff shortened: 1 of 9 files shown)


Ilya246:
may need us to make ore rarer first

OneZerooo0 (changes requested):
I’m adding more processing content so probably don’t buff material values since their value will emerge after processing

Ilya246:
> I’m adding more processing content so probably don’t buff material values since their value will emerge after processing

this should maybe also wait for us to nerf ore yield

ThanosDeGraf:
> may need us to make ore rarer first

Don't forget the materials gathered from wrecks and expeditions

Redrover1760 (changes requested):
[message(1).txt](https://github.com/user-attachments/files/27059585/message.1.txt)

98 ships fail arbitrage after this PR and are worth more than they cost.

As said in the discord, this is not the right way to fix things. Focus on making the exotic materials worth more instead of generic materials.

PR #3965, opened 2026-05-07
Blip Optimizations 1

Redrover1760:
## About the PR

Index based blip removal to speed up blip removals a bit

PhysicsQuery.TryComp instead of TryComp raw

no longer doing full hash comparisons for dictionaries, replaced with custom Equals comparison with fastest comparisons first.

minor bug with a missing inverted if check

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Client/_Mono/Radar/RadarBlipsSystem.cs
+++ b/Content.Client/_Mono/Radar/RadarBlipsSystem.cs
@@ -44,7 +44,7 @@ public sealed partial class RadarBlipsSystem : EntitySystem
     {
         for (var i = 0; i < _blips.Count; i++)
         {
-            if (_blips[i].Uid.Equals(args.NetBlipUid))
+            if (_blips[i].Uid == args.NetBlipUid)
             {
                 _blips.RemoveAt(i);
                 return;
--- a/Content.Client/_Mono/Radar/RadarBlipsSystem.cs
+++ b/Content.Client/_Mono/Radar/RadarBlipsSystem.cs
@@ -17,6 +17,7 @@ public sealed partial class RadarBlipsSystem : EntitySystem
     private static readonly TimeSpan RequestThrottle = TimeSpan.FromMilliseconds(500);
 
     private TimeSpan _lastUpdatedTime;
+    private static readonly List<HitscanNetData> EmptyHitscans = new();
     private List<BlipNetData> _blips = new();
     private List<HitscanNetData> _hitscans = new();
     private List<BlipConfig> _configPalette = new();
@@ -41,8 +42,14 @@ public sealed partial class RadarBlipsSystem : EntitySystem
 
     private void RemoveBlip(BlipRemovalEvent args)
     {
-        var blipid = _blips.FirstOrDefault(x => x.Uid == args.NetBlipUid);
-        _blips.Remove(blipid);
+        for (var i = 0; i < _blips.Count; i++)
+        {
+            if (_blips[i].Uid.Equals(args.NetBlipUid))
+            {
+                _blips.RemoveAt(i);
+                return;
+            }
+        }
     }
 
     public void RequestBlips(EntityUid console)
@@ -107,7 +114,7 @@ public sealed partial class RadarBlipsSystem : EntitySystem
     public List<HitscanNetData> GetHitscanLines()
     {
         if (_timing.CurTime.TotalSeconds - _lastUpdatedTime.TotalSeconds > BlipStaleSeconds)
-            return new();
+            return EmptyHitscans;
 
         return _hitscans;
     }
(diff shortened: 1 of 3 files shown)


Content.Client/_Mono/Radar/RadarBlipsSystem.cs
```diff
@@ -41,8 +42,14 @@ private void HandleReceiveBlips(GiveBlipsEvent ev, EntitySessionEventArgs args)
 
     private void RemoveBlip(BlipRemovalEvent args)
     {
```
Ilya246, line 44: just keep it as linq for readability, iirc modern linq is free, and besides, this is client CPU costs
we should profile this actually

Content.Shared/_Mono/Radar/RadarMessages.cs
```diff
@@ -102,4 +95,33 @@ public partial record struct BlipConfig
...
+            && RespectZoom == other.RespectZoom
+            && Rotate == other.Rotate
+            && Color == other.Color
+            && Bounds == other.Bounds;
+    }
+
+    public override readonly int GetHashCode()
+    {
+        throw new NotSupportedException("BlipConfig is not supported with GetHashCode().");
+    }
```
Ilya246, line 116: remove this once you verify it doesn't use it

Content.Shared/_Mono/Radar/RadarMessages.cs
```diff
@@ -102,4 +95,33 @@ public partial record struct BlipConfig
...
+
+    public static bool operator ==(BlipConfig left, BlipConfig right)
+    {
+        return left.Equals(right);
+    }
+
+    public static bool operator !=(BlipConfig left, BlipConfig right)
+    {
+        return !(left == right);
+    }
```
Ilya246, line 126: are those needed

Content.Shared/_Mono/Radar/RadarMessages.cs
```diff
@@ -78,13 +70,14 @@ public record struct BlipNetData
     Angle Rotation,
     ushort ConfigIndex,
     ushort? OnGridConfigIndex
+
```
Ilya246, line 73: excess

Content.Server/_Mono/Radar/RadarBlipSystem.cs
```diff
@@ -135,39 +133,35 @@ private void AssembleBlipsReport(EntityUid uid, List<EntityUid> sources, RadarCo
     /// </summary>
     private ushort GetOrAddConfig(BlipConfig config)
     {
-        if (_paletteIndex.TryGetValue(config, out var index))
-            return index;
+        for (var i = 0; i < _tempPaletteCache.Count; i++)
+        {
+            if (_tempPaletteCache[i] == config)
+                return (ushort)i;
+        }
```
Ilya246, line 140: are you sure this is actually faster

Content.Shared/_Mono/Radar/RadarMessages.cs
Redrover1760: Seems to be implemented similar to other things + IDE also told me so

Content.Shared/_Mono/Radar/RadarMessages.cs
Redrover1760: My IDE was warning me that it should have a GetHashCode() override for consistency

Content.Client/_Mono/Radar/RadarBlipsSystem.cs
Redrover1760: RemoveAt is O(1) while Remove is O(n). That's the main optimization

Ilya246: does it really not let you remove first occurence in one line

Redrover1760: Dont think so. I wrote the initial code funnily enough, I think I would have found a method to do so when I researched it before smh

Content.Server/_Mono/Radar/RadarBlipSystem.cs
Redrover1760: Doing research I think its better to implement a GetHashCode override and leave this code as is. It seems kind of fun to do so.

I need to manually perf check this to see if the difference is significant enough to warrant it yes fair as well.

Redrover1760: With the Equals override it should be able check faster versus the default record struct gethashcode since the amount of unique blip configs aren't that large. Probably.

Content.Client/_Mono/Radar/RadarBlipsSystem.cs
Redrover1760: > does it really not let you remove first occurence in one line

https://stackoverflow.com/questions/4075340/finding-first-index-of-element-that-matches-a-condition-using-linq

PR #3960, opened 2026-05-06
Draupnir Nerf

OneZerooo0:
## About the PR

Makes the draupnir fire in 25-round bursts with a 3 second delay between bursts.
## Why / Balance

Keeps the draupnir as a holy-shit-thats-a-minigun feel without it being overpowered. (Previously it basically annihilated everything in range.)
## Media

https://github.com/user-attachments/assets/58ccf849-eecc-4bb8-999c-4bd4006d5283

## Requirements

- [x] I have read relevant guidelines/documentation to this PR found on [our devwiki](https://monolith-station.github.io/mono-docs/).
- [x] I have added media to this PR or it does not require an ingame showcase.
- [x] I can confirm this PR contains either no AI-generated content, or AI-generated content that meets our guidelines.

## Breaking changes

**Changelog**

:cl: Onezero0
- tweak: Nerfed the Draupnir; It now fires in 25-round bursts, with a 3 second cooldown between bursts.

PR #3973, opened 2026-05-09
Revert "Draupnir Nerf (#3960)"

Ilya246:
## About the PR
i don't think it was at all needed, the changes kinda make it useless
draupnir was already okay, if you consider that it has terrible accuracy and range
plus the test shown in the PR showcases it being used in perfect conditions and still struggling; also, 4 of any other, even small, shipgun firing at one spot will shred it
i don't think draupnir at all needs a nerf, it's okay considering its hitrate and general spread in combat
however, there is an issue, but the issue is that mappers overuse draupnir, not that it's too strong

**Changelog**

:cl:
- tweak: Reverted draupnir nerf.


OneZerooo0 (approved):
mb gang

PR #3985, opened 2026-05-11
PDV Garm - T1 Ship [Replaces Hourglass]

lexaSvarshik:
## About the PR

New T1 PDV ship! Meet bomb-focused ship with some support weaponry (Bofors) and personal defense armament (two dravons and fang)
 
## Why / Balance

Hourglass outdated, bad weapons, bad mobility.

## Media

a bit outdated, added station record console on the bridge, other things untouched

Garm-0
Render
изображение
Light
изображение
Ship signature

Hourglass:
изображение
old pic but whatever
## Requirements

## How to test

Load, buy Garm voucher, acquire ship, look around.

## Breaking changes

**Changelog**

:cl: Svarshik
- add: Added T1 PDV ship - "WEN Garm", with GPOL bombing armament.
- remove: Hourglass-type hull no longer being produced.


lexaSvarshik:
btw gpols need fix - they just collide with eachother and explode ;-;

burrrkeye:
GPOLs will prove useless in a ship like this due massive hull and unlikeliness to actually land any shot at targets unless only advanced munition types are used and you need to reload them one by one by hand instead of with magazines in the case of vespera and vaynks and at that point just swap out for vanyks/vespera/torpedoes

Thrusters adjacent to diagonal grilles are not compliant of the guidelines

 interior work is fire btw

uhh editing again whoops, you would prefer actual point defense like scattercannons and l85 or draupnirs instead of dravons and fangs

lexaSvarshik:
Well then just buff gpols. (may do myself soon)
It is a support ship, so it should work great against slow/stationary targets, area denial. It can function as long-range artillery or use fang and dravon when out of bombs.

>Thrusters adjacent to diagonal grilles are not compliant of the guidelines

There is nothing said about that

burrrkeye:
GPOLs will underperform not because they are weak but because their purpose is to be used by fighters who can reach high speeds in reasonable times to make proper use of the 50kg that DOES NOT ACCERELATE or have a proper velocity speed of its own and COMPLETELY RELIES on ship velocity to reach its target

RemervKeberb:
I'm all for a big bomber even if it would be hard to effectively use.... 
But it needs a few things to have a chance to be viable in an actual fight. Bomb bays need a blast door and munitions bay for a loader to be able to risk his life opening them up and combat reloading without leaving the ship (the bomb bays themselves could also be given more of a protective shroud to help with this?)
Currently it has better backwards thrust than forwards? Which could be an interesting gimmick on a ship with the exact opposite play style of this ones intended tactics. Those back two small ones need to be turned into large ones at a minimum to be viable for a bomber.

Right now you have, three bomb bays. Which is one more than dedicated fighter bomber craft, and might not be entirely worthy for the sacrifices of extra bulk. Replacing a dravon with an extra bay so you have 4 would at least make it a considerable bump in damage potential worthy of the trade in lowered hitrate of the hull, the other autocannons could then be replaced with cheaper PD.

Alternatively you could just go the torpedo launcher but that's boring and honestly this kinda layout and design is really cool, good mapping work on that.

waphun:
what is against thruster guidelines though is having large thrusters facing backwards while also having one for strafing as medium ships are only allowed to have large thrusters either facing forwards/backwards or for strafing but not a mix of two.

Corrupt142 (changes requested):
Image

Reduce it to 2 guns max

Image

Guide on how to link it all up, store the logic boards somewhere in a locker 

Image

Catwalks under thrusters, dont leave em bare!

Corrupt142:
ok

lexaSvarshik:
yay

PR #4011, opened 2026-05-16
Mining Pulser Fix

Corrupt142:
## About the PR
 fixed pulser not working (had 3K starting charge but only could have 2250 max)

:cl:

- fix: Fixed mining pulser not working


Corrupt142:
goidamerge

PR #4019, opened 2026-05-18
Lathe Recipe Slop

kasature90:
## About the PR
I went through all the lathe recipes I could find and added copper/lithium/plutonium where applicable. Also did some other changes like lowering durathread cost for hardsuits because it was Really annoying and probably some other stuff I forgot.

## Why / Balance
New mats having a use outside of the voucher gambling, cool!

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Went through every lathe recipe to add Copper/Lithium/Plutonium where applicable.
- tweak: Lowered hardsuit durathread cost a bit.


OneZerooo0 (changes requested):
plutonium use is cool but I feel like some of these are not necessary (BoH and microreactor is fine, the others feel kind of lame)
You add a lot of resource requirements but you should probably reduce some other costs to account for added copper/lithium

kasature90:
> you should probably reduce some other costs to account for added copper/lithium

Done for the ones that had cost majorly affected, leaving it as is for the ones where the additional cost is only like 1.5 sheets

Ilya246:
i'm not sure about changing material costs for anything honestly, currently costs for most stuff are just material soup that doesn't matter because you have no agency in what materials you get and there's no real difference in their usecases either
we should somehow seek to improve this

kasature90:
> i'm not sure about changing material costs for anything honestly, currently costs for most stuff are just material soup that doesn't matter because you have no agency in what materials you get and there's no real difference in their usecases either we should somehow seek to improve this

It was in wanted changes so its a whateverburger right now

kasature90:
Im too lazy to keep this updated

PR #4021, opened 2026-05-19
Change Tax Into Savings

Ilya246:
## About the PR
now instead of vanishing money goes into savings
savings can be used for loadout
replaces monocoins
fixes "TSF Central Bank"

also cleans up code

## Media
image

after a few (note that balance is broken in this particular screenshot)
image

**Changelog**

:cl:
- add: Instead of being taxed, excess money now goes to your savings account, which can be used for loadouts (and possibly more in the future)
- remove: Taxes.
- remove: Monocoins.
- fix: No more "TSF Central Bank".


Ilya246:
evil selfmerge....

PR #4037, opened 2026-05-21
Fix Cortical Borers Burning to Death when Inside Host that Is in Superheated Environment.

SarahRaven:
## About the PR

Modified TemperatureSystem.cs to add in a check for HasImmunity, similar to the PressureImmunityComponent added to the BarotraumaSystem.cs

Also added the TemperatureImmunityComponent to the OnInfest of the Cortical Borer to apply and remove when in play.

## Why / Balance

I found it quite silly how the Cortical Borer could be within a 100% heat resistant hardsuited person in a tritium fire, and the Cortical Borer would die of heat stacks and burning to death while the person they infest is perfectly safe from the heat and fire.

## Media

image
image
image

## Requirements

## How to test

1. Spawn a cortical borer.
2. Enter a host.
3. Have the host wear a 100% heat resistant hardsuit (Like CE Daedalus suit, Aeolus modsuit, or Atmos suits).
4. Set fire to your room (tritium fire, or welding fuel fire will work)
5. Watch as host is completely engulfed in fire, room is over 100C, yet borer is alive and comfy.
6. Leave host.
7. Watch as cortical borer is completely engulfed in fire, room is over 100C, and borer is now BBQ.

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Body/Systems/ThermalRegulatorSystem.cs
+++ b/Content.Server/Body/Systems/ThermalRegulatorSystem.cs
@@ -54,16 +54,14 @@ public sealed partial class ThermalRegulatorSystem : EntitySystem
             return;
 
         // mono begin
+        // Check to see if we're disabling thermal temporarily
+        if (ent.Comp1.DisableProcessing)
+            return;
+
         if (ent.Comp1.ProcessWhileDead == false && TryComp<MobStateComponent>(ent, out var mobComp1) && mobComp1.CurrentState == MobState.Dead)
             return;
 
         if (ent.Comp1.ProcessWhileCrit == false && TryComp<MobStateComponent>(ent, out var mobComp2) && mobComp2.CurrentState == MobState.Critical)
-            return;
-
-        // Check to see if we're disabling thermal temporarily
-        var disableThermal = new CheckSkipRegulationEvent();
-        RaiseLocalEvent(ent, ref disableThermal);
-        if (disableThermal.SkipRegulation)
             return;
         // mono end
 
--- a/Content.Server/Body/Systems/ThermalRegulatorSystem.cs
+++ b/Content.Server/Body/Systems/ThermalRegulatorSystem.cs
@@ -59,6 +59,12 @@ public sealed partial class ThermalRegulatorSystem : EntitySystem
 
         if (ent.Comp1.ProcessWhileCrit == false && TryComp<MobStateComponent>(ent, out var mobComp2) && mobComp2.CurrentState == MobState.Critical)
             return;
+
+        // Check to see if we're disabling thermal temporarily
+        var disableThermal = new CheckSkipRegulationEvent();
+        RaiseLocalEvent(ent, ref disableThermal);
+        if (disableThermal.SkipRegulation)
+            return;
         // mono end
 
         var totalMetabolismTempChange = ent.Comp1.MetabolismHeat - ent.Comp1.RadiatedHeat;
--- a/Content.Server/Atmos/EntitySystems/FlammableSystem.cs
+++ b/Content.Server/Atmos/EntitySystems/FlammableSystem.cs
@@ -110,9 +110,6 @@ namespace Content.Server.Atmos.EntitySystems
             }
         }
 
-        // Mono: TryComp to see whether target is immune to heat stacks
-        record struct TryIgniteEvent(bool Cannceled = false);
-
         // Frontier: ignition on projectile hit event
         private void OnProjectileHit(EntityUid uid, IgniteOnProjectileHitComponent component, ProjectileHitEvent args)
         {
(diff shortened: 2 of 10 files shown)


SarahRaven:
(For the last picture, I took it too early. The borer indeed burned to death once the heat kicked in after leaving the host)

Ilya246 (changes requested):
i'll have to ask you to redesign how this works: remove all upstream changes, subscribe cortical borers to ModifyChangedTemperatureEvent and set TemperatureDelta on it to 0 if we're inside a host 
maybe add `heatAmount == 0` check into TemperatureSystem that does an early `return`
also make a `record struct TryIgniteEvent(bool Cancelled = false);` somewhere near FlammableSystem and replace your TryComp with firing that event and doing a `return` if cancelled

SarahRaven:
@Ilya246 

For some reason when I subscribe to the ModifyChangedTemperatureEvent in CorticalBorerSystem.cs with:
```
         SubscribeLocalEvent<CorticalBorerComponent, ModifyChangedTemperatureEvent>(OnTemperatureChange);
```
and:
```
     private void OnTemperatureChange(Entity<CorticalBorerComponent> ent, ref ModifyChangedTemperatureEvent args)
     {
         // Affected by heat outside of host. In future, could check to synchronize with heat stacks and temp of hardsuit'd host.
         if (!ent.Comp.Host.HasValue)
             return;

         Console.WriteLine("TemperatureDelta is {0}", args.TemperatureDelta);
         args.TemperatureDelta = 0f;
     }
```

I notice that the TemperatureDelta always ends up being something like -870f for 20C, and goes insanely high/low (something like -30000 to 60000) when on fire. The weirdest part is when I force set TemperatureDelta to zero, that somehow causes the Cortical Borer to start overheating, to the point of heat death about 30-60 seconds later.

I'm not sure what I'm doing wrong with it.

Ilya246:
log its temperature from temperature component and see what it does
also, the delta being large is probably because it's actually heat and not temperature, aka misnamed variable

SarahRaven:
Found the issue. Seems that CorticalBorers will calculate their heat with ThermalRegulation component, so after you set the TemperatureDelta [Heat] to zero, it will try to heat up, cascading until overheating. We will likely have to disable thermal regulation while inside the body too.

Ilya246:
does that heat you up in bypass of the event? i'd say just add a bool to thermal regulation that enables/disables it and switch it on enter/leave host

SarahRaven:
Yes, it heats them up in bypass of the event. I made it so you could disable ThermalRegulation temporarily with a raised local event and replying to it in CorticalBorer system to reply whether they're in a host or not.

SarahRaven:
It's just really weird how ThermalRegulation adds heat and completely skips over the Heat events of the Temperature system.

Sometimes I wonder where the fuck the code goes, and we find 'this'. I've done the change, still testing, but it does seem to avoid heating up and burning while inside the host. The ONE big issue currently is that if you have heat stacks and are burning before going into a host, it'll set your HEAT to normal with TemperatureDelta = 0, but the ignite stacks trigger before they can be removed, thus the Borer's temperature will skyrocket inside. Since we force TemperatureDelta to zero, that means no heat will be removed/added, thus they will still burn.

Content.Server/Body/Systems/ThermalRegulatorSystem.cs
```diff
@@ -59,6 +59,12 @@ private void ProcessThermalRegulation(Entity<ThermalRegulatorComponent, Temperat
 
         if (ent.Comp1.ProcessWhileCrit == false && TryComp<MobStateComponent>(ent, out var mobComp2) && mobComp2.CurrentState == MobState.Critical)
             return;
+
+        // Check to see if we're disabling thermal temporarily
+        var disableThermal = new CheckSkipRegulationEvent();
+        RaiseLocalEvent(ent, ref disableThermal);
+        if (disableThermal.SkipRegulation)
+            return;
```
Ilya246, line 65: should just be a bool instead like i said for performance reasons

SarahRaven:
Looks like it also keeps your heat before you go in. Wish I could just activate what we use for the OnRejuvenate where it forces TemperatureComponent's to Atmospherics.T20C, but yeah.

Content.Server/Atmos/Components/FlammableComponent.cs
```diff
@@ -83,4 +83,12 @@ public sealed partial class FlammableComponent : Component
...
         public ProtoId<AlertPrototype> FireAlert = "Fire";
     }
+
+    // Mono: Event to see whether target is immune to heat stacks
+    /// <summary>
+    /// Event raised on an entity to determine if it is immune to ignition. 
+    /// </summary>
+    [ByRefEvent]
+    public record struct TryIgniteEvent(bool Cancelled = false);
+
```
Ilya246: standard is to put this inside the system or in its own file

SarahRaven: Understood, moving it to Content.Server/Atmos/EntitySystems/FlammableSystems.cs (end of namespace).

Content.Server/Body/Systems/ThermalRegulatorSystem.cs
SarahRaven: Understood. It's now toggled OnInfest and EjectHost. Sadly this required removing the access control from Content.Server/Body/Components/ThermalRegulatorComponent.cs because it was complaining about Read-only access to the variable due to `[Access(typeof(ThermalRegulatorSystem))]`.

SarahRaven:
Going to rebase to try and fix the FlammableComponent change

Ilya246 (approved):

PR #4043, opened 2026-05-22
Please Just Fly Normal Ships

RemervKeberb:
## About the PR
Makes Ultra-High-Gunnery servers are no longer unanchorable or deconstruct-able.
Omega server GP power down from 500 --> 200

## Why / Balance
Believe it or not, we can do game balance better than this. 
image
image
Saturn has 162 gp
Flyssa has 180 gp

180 is already double that of an ultra-high gunnery server which is now no longer going to be the norm, 200 lets the existing capitals have some wiggleroom for modding but realistically, the base armaments of the capital ships should just be enough on their own mapping wise.

As for the ultra-high server...
The current state of affairs for ship combat is rather bleak as players homogenize every single ship design into the same brick with 90gp worth of guns they spent an hour building instead of playing the game. When they finally do play the game, it either trivializes fights against others who do not rush the same power-gaming time sink as them making the fight unfun for both parties. In the event they DO meet their match, both pilots end up flying with an unreasonable level of fear as they do not want their real life time investment of the past hour to suddenly go up in flames, leading to these upgraded ship fights being avoided and ran from more often than stock ship fights, leading to more unsatisfying engagements and outcomes.

Ultra-high gunnery servers are not researchable so it stands to reason that they should not be so easily acquired. Instead of simply buying and selling a ship to acquire, they can be rare drops at POI's or rewards for engaging in ship combat in the first place, literally anything would be better than this, including nothing. Which is the stopgap measure I'm offering here.

## Requirements

## How to test
Spawn ultra-high-gunnery server, try to unwrench or deconstruct. 
Spawn flyssa or saturn, observe that all their guns still work and can house an extra handful of guns but not enough to blot out the sun.

**Changelog**
:cl:
- tweak: Ultra-high-gunnery servers no longer unanchorable.
- tweak: Omega-high-gunnery server GP down from 500 --> 200


RemervKeberb:
Whoops. That's the wrong commit line.

lexaSvarshik:
btw wywern has ~280 power

ari771155:
i honestly wish ship modding was either non present or significantly harder / restricted - so i'm all for this

PR #4066, opened 2026-05-27
Added Leporazine, Inaprovaline, and Pax to Cortical Borer Chems

SarahRaven:
## About the PR

Added a few more chemicals (Leporazine, Inaprovaline, Pax, and others) to the Cortical Borer's internal process. Also increased the starting window height of the BoundUserInterface for the Cortical Borer chemical window to accommodate.

## Why / Balance

This would allow Cortical Borer's to use inaprovaline to reduce bleeding and bring their host out of critical state faster (alongside Dexalin Plus). Leporazine is great if their host is rushing through space and they need one more moment of safety from the cold.

Pax was added as one more toy to their arsenal in case they need to escape their host. 

## Media

image
image

## Requirements

## How to test

1. Spawn a cortical borer.
2. Take control of borer and infest a humanoid with bloodstream.
3. Open chemical menu.
4. Use Inaprovaline/Leporazine/Pax.
5. Inspect bloodstream solution of host.

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Client/_Mono/CorticalBorer/CorticalBorerDispenserWindow.xaml
+++ b/Content.Client/_Mono/CorticalBorer/CorticalBorerDispenserWindow.xaml
@@ -5,7 +5,7 @@
     xmlns:ui="clr-namespace:Content.Client.Chemistry.UI"
     Title="{Loc 'reagent-dispenser-bound-user-interface-title'}"
     MinSize="400 175"
-    SetSize="600 300">
+    SetSize="600 375">
     <BoxContainer Orientation="Horizontal">
         <BoxContainer Orientation="Vertical" MinWidth="170">
             <Label Text="{Loc 'reagent-dispenser-window-amount-to-dispense-label'}" HorizontalAlignment="Center" />
--- a/Content.Client/_Mono/CorticalBorer/CorticalBorerDispenserWindow.xaml
+++ b/Content.Client/_Mono/CorticalBorer/CorticalBorerDispenserWindow.xaml
@@ -5,7 +5,7 @@
     xmlns:ui="clr-namespace:Content.Client.Chemistry.UI"
     Title="{Loc 'reagent-dispenser-bound-user-interface-title'}"
     MinSize="400 175"
-    SetSize="600 375">
+    SetSize="600 475">
     <BoxContainer Orientation="Horizontal">
         <BoxContainer Orientation="Vertical" MinWidth="170">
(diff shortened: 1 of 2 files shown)


SarahRaven:
Part of me wants to add in Barozine as well, but it seems overkill. Sure it could be used with barozine + leporazine to make someone space immune temporarily, but we're not here to JUST make mediborers.

(Still wish we could get space mirage to make actual hallucinations in-game)

lordsej:
if were going mediborer, we should atleast go full hog and go for something for rads aswell. 

Theres also possibly the more just plain fun chems like honk, weh or hew

SarahRaven:
The funniest thing with weh and hew is that if they overdose the person on it, they'll turn into a lizard, ejecting the cortical borer out of the body, which is a high risk maneuver lol.

In terms of rads, that would be interesting, but also the borer will die since they take full rads and can't chem themselves. Interesting possibility though...

lordsej:
i was thinking more a case of Arithrazine to cure rads in the host. 

Didnt think of the weh lizard transformation side effect. was there a PR at somepoint that fixed when a borer host turned into a chimera, the borer wouldnt end up dead in the either. maybe the same conversion fix works for lizardfication

SarahRaven:
Oh yeah, I think a PR was made to fix it. You know what, let's just add for fun.

SarahRaven:
Okay there is one MAJOR issue with doing both JuiceThatMakesYouHew and JuiceThatMakesYouWeh.... if you mix the two, it causes the person to explode inside and deals damage to them split across Rads/Brute/Heat. It's HELL to heal.

Might be a bit too strong, but then again heartbreaker toxin exists.

lordsej:
how flexable is the borer chem price. why not set it so every amount of injectable from 1/5/10/15/20/25/30/50/100 for both weh and hew to cost exactly 200 borer chemicals each. 

yes you can temp turn someone to a lizard but by the time you regenerate back to 200 points the chemicals should have left the body not causing a hew/weh explosion

Corrupt142:
- Give us a good reason for borers to have Leporazine and Inaprovaline other than mediborer gameplay.

- the Pax can work.

SarahRaven:
Did Leporazine lose the effect to sleep hydrakin? Leporazine can also be used on Lizards to knock some out. Have to think if it's low enough to slow them. Inaprovaline can be removed. I just like to say "keeping host alive means more eggs", which isn't a direct "only mediborer" thing to do.

In terms of Hew/Weh/Honk, guessing those too are fine? That said, it does seem we're just adding more things to annoy the host into getting rid of the reproducer.

kasature90 (approved):

PR #4124, opened 2026-06-22
Camelot Remap

Corrupt142:
## About the PR
 chuud
lowered the starting reactor pressure to roughly 600KPA

Replaced NSTRA suit with commisar parade suit storage in commisars office

webified RSC section (its currently dead af)

fixed a LV network in the top RIGHT module


Corrupt142:
err.. wrong branch 1 sec

PR #4132, opened 2026-06-24
Adjusts a Lot of Things with LPCs.

Mrrpip:
## About the PR

* Does exactly what it says on the tin. In every case, I believe it was just halving the time so 2 hours becomes 1, etc, etc.

* Adjust renewable LPCs to have 12 uses, this is separate from cooldowns as all ships could only be redeemed once in the current build.

* Lowers cooldown from unassigning an ID from 1 hour to 25 minutes. 
## Why / Balance

* LPC times weren't correct and doubled in most cases, so that's an obvious fix.

* LPCs only had one use, even when they're supposed to be cooldown based. Went with 12 as while **very** unlikely, you can get up to 11 Ledokol attachments if you get one after another from the first half hour.

* Lowered the timer, because unassigning from an LPC still incurs the character penalty of 1 hour. Should still stop ship spam, while also allowing factions to renew FUBAR'd ships more often.
## Media

Screenshot_20260624_150031
Screenshot_20260630_044849

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Renewable LPCs in the sector were found to be running under Emissions Savings Time. They now line up their intended recharge times.
- fix: Renewable LPCs can be used more than the initial purchase.
- tweak: Player deed unassign timer lowered to 25 minutes.


Mrrpip:
Screenshot_20260624_150230

A bit more context for potential future changes and should probably be kept in mind by the reviewer. For now though, we're keeping it as just fixing the mismatched times until told otherwise.

Mrrpip:
PR changed a LOT. I'd really recommend going trough it with a comb when reviewing.

NazrinNya (approved):

NazrinNya (changes requested):
actually no too evil.
make uplink LPC's way more expensive then (t1-t2 in question)

Mrrpip:
> actually no too evil. make uplink LPC's way more expensive then (t1-t2 in question)

I guess I can do that, but it raises the question did we just know LPC times were wrong and not do anything?

waphun:
> > actually no too evil. make uplink LPC's way more expensive then (t1-t2 in question)
> 
> I guess I can do that, but it raises the question did we just know LPC times were wrong and not do anything?

yes

Mrrpip:
Alright, I'm basically throwing shit at the wall here to see what sticks, so the first nudge is going to be a lot less extreme.

TSF

* Spica unchanged. It... Isn't great on the best of days.
* Most tier 1s increased by 33%, requiring dipping into the round start central FMC pot, pooling together FMC or being a captain.
* Mercury increased by 20% as the developer comment mentions needing 2 and being discounted.

* Tier 2s increased by 10%
* Tier 3+ unchanged.

PDV

* I'm not entirely certain on the Garm. I'm increasing it 33% because I'm of the opinion it'd punch way above it's tier if GPOB Quasars actually worked as intended. Can still be bought on the default starting DC amount.

* All other T1s increased by 33% and can be bought as either Rakhshan + central pot or Asvaran.

* Tier 2s increased by 10%
* Vulture unchanged. It's a mining ship.
* Tier 3s unchanged.


I do feel like this is a bit odd, but on the other hand, the three you start with aren't RNG these days.

PR #4137, opened 2026-06-25
Fix Map Painter

Ilya246:
title, makes map painter not crash itself when trying to use it


Ilya246:
selfmerging without tests because no way this breaks anything

PR #4181, opened 2026-07-02
Fix(loc): Register MANY/MAKEPLURAL for the Es-ES Culture

TheLacrox:
## Problem

Server logs were flooded with:

```
[ERRO] loc: "es-ES"/"reagent-effect-guidebook-status-effect": "Unknown function: MANY()"
```

for `reagent-effect-guidebook-status-effect`, `-movespeed-modifier`, `-electrocute` and `-reduce-rotting`.

## Root cause

`ContentLocalizationManager.cs` registers the language-specific Fluent functions (`MANY`, `MAKEPLURAL`) only on the **en-US fallback culture** (upstream comment even says other languages should add their own). The es-ES tree calls `MANY()` 8× and `MAKEPLURAL()` 11×, so the active es-ES bundle hit an unknown function every time those messages rendered.

## Changes

- `ContentLocalizationManager.cs` (the one approved Capibara divergence): register `MANY`/`MAKEPLURAL` on the es-ES culture with Spanish pluralization rules (vowel → +s, `í/ú` → +es, `z` → -z+ces, `s/x` invariable, consonant → +es). Bracketed with `// Capibara ESP` markers.
- `es-ES/guidebook/chemistry/effects.ftl`: translate the literal argument `MANY("second", …)` → `MANY("segundo", …)` (8×) so times render as "3 segundos" instead of "3 seconds".
- `es-ES/_EinsteinEngines/power/batteryDrinker.ftl`: fix `CAPATALIZE` → `CAPITALIZE` typo (copied verbatim from upstream en-US, which still has the bug — en-US left untouched per fork policy).
- `CapibaraCultureTest`: regression assertions that `MANY()` resolves with Spanish singular/plural inside the es-ES bundle, via new seed key `capibara-loc-many`.

## Verification

- `pwsh _Capibara/validate-locale.ps1` → exit 0, "All es-ES files structurally valid."
- `dotnet test Content.IntegrationTests --filter CapibaraCultureTest` → 1 passed, 0 failed.

## Note: untranslated "Inner Ring" (out of scope)

The location splash texts ("Inner Ring", "Civilised space…") come from `Resources/Prototypes/_Crescent/Biomes/space_biomes.yml` (`ambientSpaceBiome` prototypes). `SpaceTextDisplaySystem` draws `biome.Name`/`biome.Description` **verbatim — there is no Loc call**, so they cannot be translated additively. Fixing them requires either a new (tiny) C# divergence in `Content.Client/_Crescent/SpaceBiomes/SpaceBiomeTextDisplaySystem.cs` (`Loc.TryGetString($"space-biome-{id}-name")`-style lookup) or translating the YAML in place like the ServerInfo exception. Left out of this PR pending a policy decision.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


TheLacrox:
Opened against the wrong repo (this is a downstream translation fork change) — sorry for the noise.

PR #4239, opened 2026-07-07
Fix 4 Letter Ship Designations Being Fully Renamable

kasature90:
image

:cl:
- fix: Fixed ships with 4 letter designation suffixes (such as TSFN/USSP) being fully renamable.


kasature90:
selfmerging on checks pass

PR #4240, opened 2026-07-07
Chatranks

kasature90:
Also adds support for loadout prototypes to add comps to players.
image
image

:cl:
- add: Added Chatranks for CC, MD, VG, USSP, TSF, and PDV. TSF gets selectable chatranks in loadout, others don't because I'm lazy and nobody made full ranks for them yet.

### DIFF
--- a/Content.Server/_Crescent/Chat/ChatRankComponent.cs
+++ b/Content.Server/_Crescent/Chat/ChatRankComponent.cs
@@ -9,5 +9,5 @@ namespace Content.Server.Crescent.Chat;
 public sealed partial class ChatRankComponent : Component
 {
     [DataField]
-    public string Rank;
+    public LocId Rank;
 }
--- a/Content.Server/_Crescent/Chat/ChatRankComponent.cs
+++ b/Content.Server/_Crescent/Chat/ChatRankComponent.cs
@@ -1,0 +1,13 @@
+namespace Content.Server.Crescent.Chat;
+
+/// <summary>
+/// Stores "rank" of the user for certain contexts.
+/// Mostly given by role.
+/// </summary>
+
+[RegisterComponent]
+public sealed partial class ChatRankComponent : Component
+{
+    [DataField]
+    public string Rank;
+}
--- a/Content.Shared/Station/SharedStationSpawningSystem.cs
+++ b/Content.Shared/Station/SharedStationSpawningSystem.cs
@@ -130,6 +130,10 @@ public abstract partial class SharedStationSpawningSystem : EntitySystem
     {
         EquipStartingGear(entity, loadout.StartingGear, raiseEvent);
         EquipStartingGear(entity, (IEquipmentLoadout) loadout, raiseEvent);
+
+        // Mono - loadout comps
+        var entMan = IoCManager.Resolve<IEntityManager>();
+        entMan.AddComponents(entity, loadout.Components);
     }
 
     /// <summary>
--- a/Content.Shared/Station/SharedStationSpawningSystem.cs
+++ b/Content.Shared/Station/SharedStationSpawningSystem.cs
@@ -132,8 +132,7 @@ public abstract partial class SharedStationSpawningSystem : EntitySystem
         EquipStartingGear(entity, (IEquipmentLoadout) loadout, raiseEvent);
 
         // Mono - loadout comps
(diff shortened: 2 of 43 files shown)


AthrunZala33092:
As much as I like this concept, I want to be able to choose the rank to spawn with, or I'm not going to be very happy with this. This doesn't account for special ranks that some people use (Like WO Warrant Officer). I'm hoping that this isn't just throw Corporal and Sergeant at the ranks and call it done. We have Privates, Corporals, Sergeants, Staff Sergeants, Majors, Lt. Colonels, etc. Also Sergeant is not Srg, it's Sgt.

Zethine07:
Honestly could work if you could set honorifics in loadout rather than making it a forced thing. Like a little alternative tag thing for each faction you can type in or select for a rank.

Technically this already exists in the form of the loadout names but that changes the whole name so like yeah.

Meroulkas:
funky as a system where the same job have diferent pre made title, for exemple security officer can be something like peace keeper, or night watch, could maybe take that systeme if this is what you looking for

kasature90:
> As much as I like this concept, I want to be able to choose the rank to spawn with,

> Honestly could work if you could set honorifics in loadout

image

EldritchPineapple:
hm yes i would like to order a mc colonel for 0 dollar

AthrunZala33092:
If I'm reading this file right and the Major O-4 has been moved to the Captain loadout pick, it's done. The engineers have flair. The corpsman have flair. This is truly peak. 😭

Content.Server/_Crescent/Chat/ChatRankComponent.cs
```diff
@@ -0,0 +1,13 @@
...
+/// <summary>
+/// Stores "rank" of the user for certain contexts.
+/// Mostly given by role.
+/// </summary>
+
+[RegisterComponent]
+public sealed partial class ChatRankComponent : Component
+{
+    [DataField]
+    public string Rank;
```
Ilya246: LocId

Content.Shared/Station/SharedStationSpawningSystem.cs
```diff
@@ -130,6 +130,10 @@ public void EquipStartingGear(EntityUid entity, LoadoutPrototype loadout, bool r
     {
         EquipStartingGear(entity, loadout.StartingGear, raiseEvent);
         EquipStartingGear(entity, (IEquipmentLoadout) loadout, raiseEvent);
+
+        // Mono - loadout comps
+        var entMan = IoCManager.Resolve<IEntityManager>();
+        entMan.AddComponents(entity, loadout.Components);
```
Ilya246: don't use ioc, EntityManager.AddComponents(...);

Resources/Locale/en-US/_Mono/rank/ranks.ftl
```diff
@@ -0,0 +1,69 @@
...
+tsfn-e3-cn = HJ3-ZW
+tsfn-e4-cn = HJ4-SW
+tsfn-e5-cn = HJ5-SHX
+tsfn-e6-cn = HJ6-ZX
+tsfn-e7-cn = HJ7-SX
+tsfn-e8-cn = HJ8-DX
+tsfn-e9-cn = HJ9-SHJ
+
+### PDV ranks
+## Spaska uses no chatrank so they can hide obviously. It doesnt mean much anyways with voicemask lol
```
Ilya246, line 48: maybe wearing a voicemask in general should prevent chatrank from showing up

Ilya246 (changes requested):

Resources/Locale/en-US/_Mono/rank/ranks.ftl
kasature90: requires more effort than i'm willing to put in for things nobody uses anyways

Content.Shared/Station/SharedStationSpawningSystem.cs
kasature90: fixed

Content.Server/_Crescent/Chat/ChatRankComponent.cs
kasature90: fixed

Ilya246 (approved):

PR #4296, opened 2026-07-16
Grshitfix

eris-webserv:
I'm not even giving this PR a proper description; it doesn't deserve it. This fixes the gravity of shipguns when unanchored in space. That's it.

Your contributors are all bums who can't get off their asses to make a single line YML fix and instead complain about it.


kasature90 (approved):

eris-webserv:
If you're wondering why this PR also adds an extra newline in base_structure, I added it there before realizing it was unneccesary (I'm blind), removed it, noticed the newline, and Git refuses to aknowledge that I removed it and let me push the commit where I did.

PR #4322, opened 2026-07-20
Test

kasature90:
ark told me in a divine message to test doing this to fix discord CL since router doesnt use CDN

:cl:
- fix: Fixed something somewhere, it's just a test.


kasature90:
Nevermind, I'll just fix it with router the normal way

PR #4344, opened 2026-07-22
Shenzhen/Neptune Nerfs and Plates Fix

AthrunZala33092:
## About the PR
I'm going to ENJOY the public discourse on this one.

Neptune: I know that this thing has been rampantly deployed ever since T1 vouchers stopped being gatcha and the PDV uses it because it's cheap and it's the best lower tier ship they got. Why use their other even higher tier ships if they can spam Neptune and throw ms-250 shield on it to win? Bonus if you use it to ram. I know there are plans to shield shunt it, but right now, that is a while off, the Neptune map still has a shield on it...and right now in my opinion it functions much like the former Andromeda which was ALSO spammed heavily. It's a sub-tier 2 ship masquerading as a tier 1 ship, and this will force the PDV to choose what kind of T2 ships they want instead of going 'Neptune, duh'. Once the other changes the maints want to do to PDV tech in general come through, it can be re-evaluated where it belongs in the Tier list.

The SHENZHEN: Such a hot topic of debate for the PDV, this thing feels like it was made to counter the juggsuit. It used to do 120 damage with AP modification and it would guarantee a 1 hit, 1 shot critical, or stamcrit if a plate blocked it. It always felt weird to me that a laser coilgun that goes through glass doors shoots a pierce projectile anyway. I have modified the stats to do TRUE damage of 55 heat/5 shock/20 pierce. That way it doesn't have to overcompensate for a high AP juggsuit and overdamages everything in its path. It also allows the PDV to use either a heat or pierce plate to try to negate some of the damage. The planned PDV suit modifications to heat resistance is going to have no bearing on this since it's doing true damage.

Now for the PLATES: DTDumb and I were going over the code and we discovered that the 1.2 pierce resistance didn't add extra pierce damage as intended, it actually negated it. Same went for heat: if people had been running heat plates, they would've been resistant to both heat AND pierce, just nobody was using it since lasers weren't exactly super common. And likewise, SPEED plates pretty much negated everything. It's been modified to do -0.2 damage, which math wise..applies the 20% extra damage properly now. (Also in some places heat was incorrectly labeled as burn.)

## Media

Screenshot 2170

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Moved Neptune to T2
- tweak: Modified Shenzhen to not be a 1 shot gun
- fix: Fixed plates to properly add damage it's weak to

### DIFF
--- a/Resources/Prototypes/_Mono/Entities/Clothing/OuterClothing/Armor/armor_plates.yml
+++ b/Resources/Prototypes/_Mono/Entities/Clothing/OuterClothing/Armor/armor_plates.yml
@@ -19,7 +19,7 @@
     absorptionRatios:
       Blunt: 0.7
       Slash: 0.7
-      Burn: 1.2
+      Heat: -0.2
   - type: Damageable
     damageContainer: Inorganic
   - type: Destructible
@@ -50,7 +50,7 @@
     staminaDamageMultiplier: 1.0
     absorptionRatios:
       Piercing: 0.7
-      Burn: 1.2
+      Heat: -0.2
   - type: Damageable
     damageContainer: Inorganic
   - type: Destructible
@@ -81,7 +81,7 @@
     staminaDamageMultiplier: 1.0
     absorptionRatios:
       Heat: 0.7
-      Piercing: 1.2
+      Piercing: -0.2
   - type: Damageable
     damageContainer: Inorganic
   - type: Destructible
@@ -111,10 +111,10 @@
     sprintSpeedModifier: 1.1
     staminaDamageMultiplier: 2
     absorptionRatios:
-      Burn: 1.2
-      Piercing: 1.2
-      Blunt: 1.2
-      Slash: 1.2
+      Heat: -0.2
+      Piercing: -0.2
+      Blunt: -0.2
+      Slash: -0.2
   - type: Damageable
     damageContainer: Inorganic
   - type: Destructible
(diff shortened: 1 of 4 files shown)


ErolThurn:
Neptune is not worth fissile in hyperwar, its actually not even worth fissile in normal rounds if you dont heavily upgrade it , moving it up a whole tier up to things like the ganymede and europa is slop because the rest of pdv t1s are shitboxes that dont compare to tsf ships. Plus on hyperwar PDV cant even get shields on them (moving shields from t3 ships isn't allowed). This is a real slop idea ngl....

waphun:
The Shenzhen can't go through glass stuff any more, that was an oversight with the original iteration of it because it's not a laser, it's supposed to shoot a projectile that is so fast, compared to other ballistics, that it hitscans, kinda like the ADS's Flayer auto-coil.

OneZerooo0 (changes requested):
this is why I made gatcha LPCs
Neptune shouldn’t be t2, though I do agree with nerfing it
Shenzen is a coilgun, not a laser - I like the idea of having both types of plates be viable for it + full AP with no oneshot capability though
Plates are good, thought I fixed the burn references earlier…

kasature90:
these problems are already being addressed elsewhere, nothing i can say to do for neptunes right now but you can still counter a shenzhen right now with flashes/smoke/teargas (hides from therms) etc; it has an obvious laser and if you walk into that you shouldnt get surprised by getting killed

PR #4355, opened 2026-07-24
Vox Hair Additions and Marking Logic Fix (Wizden #38906)

Luweth:
## About the PR
Partial cherry-pick of https://github.com/space-wizards/space-station-14/pull/38906
Much of the original PR is already implemented in other ways so I did not cherry-pick the entire thing.
Adjusts marking limits for Vox. Limit numbers based on various other species in the game already.

## Why / Balance
More hair is a good thing, these additions along with the limit changes should allow for much more unique and interesting looking Vox to be possible.

**As for the C# change:**
Having more points than default markings, (like 6, for example as a vox) caused the while loop in ensureDefault to keep running and exceed the list, throwing an error. By changing the logic from || to && the loop only continues if both conditions are true, instead of if only one condition is true, preventing this error.
This SHOULD in theory allow more races to have more points too, if that's something people want to do in the future.
(This change has been implemented by other code bases without issue, so it should not be a problem here.)

## Media
image
image
image
image

## Requirements

## How to test
Vox character creation, yaya.

## Breaking changes
None found in testing.

**Changelog**
:cl:
- add: Added four new Vox quilstyles.
- tweak: Changed making limits for Vox.


OneZerooo0 (approved):

kasature90 (approved):
Big news for the 2 people who play vox here

PR #4376, opened 2026-07-28
Kill Simplemobs From Wrecks

kasature90:
free lag tutorial

:cl:
- remove: Removed mobs from blue wrecks. It's a performance thing, sorry.


kasature90:
i've been notified the lag from this can be fixed in a better way

PR #4378, opened 2026-07-29
TSF Laser Sentry Turret

AthrunZala33092:
## About the PR

So I think I did this right in creating the laser object, though I don't understand why it isn't using the second mode for the yellow/stun mode. (Not that anyone would ever set it to stun mode anyway.) I would've created a disabler hitscan but I was struggling to create a new entity that stamcrits right, and the turret seems incapable of swapping between types anyway. (The stun mode for the ballistic turrets *sort of* fire rubbers but also do a LOT of damage, so they seem borked anyway?)

Anyway, TSF gets a remote version of their laser turret to hook up to the turret control panel.

## Requirements

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/_Mono/Entities/Structures/Wallmounts/turret_controls.yml
+++ b/Resources/Prototypes/_Mono/Entities/Structures/Wallmounts/turret_controls.yml
@@ -57,6 +57,65 @@
     - AllHostile
 
 - type: entity
+  parent: WeaponTurretLaserTSFMC
+  id: WeaponTurretLaserBase
+  name: remote laser sentry turret
+  abstract: true
+  components:
+  - type: AccessReader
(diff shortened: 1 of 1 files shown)


kasature90 (changes requested):
this needs to work w/ the firemodes properly

AthrunZala33092:
> this needs to work w/ the firemodes properly

I was hoping you'd be able to help me with that. Despite specifying the firemodes, it doesn't actually switch when you use stun/lethal. (Though to be very fair, who uses stun ANYWAY?) I found that while the rubbers were used by the ballistic turrets, it still seems to be doing way too much damage, like it's overlaying some normal bullets on top. So it seemed to me like the sentry turret system itself was borked.

But I could be wrong?

PR #4381, opened 2026-07-29
Alt Iridite Process

kasature90:
## About the PR
Adds an alternate process for iridite production via arc furnace/centrifuge/industrial ore processor and a lot of reagents. Really expensive. Also reduces the value of iridite from 1250/sheet to 300/sheet.
Quite simple, just a bit tedious w/ chems and the recipe times I guess. Maybe it's a good way to help people learn the economy devices before Industry comes out.

## Why / Balance
Crystallizer is boring, and not very accessible (it needs a lot of space and a lot of work for it; basically nobody has that), and its better for the idea of Mass Producing 20 Crates of Iridite but it's really bad if I want to use this in any recipes ever. Now I have a 2nd annoying to get material to work with in recipes instead of fissile. But I probably won't use it for anything until arc furnaces are more obtainable instead of being MMC exclusive.
Also, I can semi-automate this one with goobtorio arms and stuff and thats cooler than 2 centrifuges on loop.

## Media
image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added an alternate recipe for Iridite that doesn't need atmos magic. Now all you need is an industrial arc furnace, industrial ore processor, a processing centrifuge, and a chemlab (with a good chemist).
- tweak: Lowered the price of Iridite to 300/sheet instead of 1250/sheet.


OneZerooo0 (approved):
Don’t think it’s price should be nuked because the recipe for 1 sheet is already cbt

Otherwise kino

kasature90:
> Don’t think it’s price should be nuked because the recipe for 1 sheet is already cbt

You do the atmos way that does like 5/sec if you want to make money, you do this one if you want to fabricate with it

PR #4388, opened 2026-07-29
Why So Sloppy? (Random Portstrike)

OneZerooo0:
## About the PR

Randomizes portstrike announcement period to happen from 1-3 hours in roundtime.
## Why / Balance

Constant preparations around portstrike at the 4 hour mark is bland.
## Requirements

## Breaking changes

**Changelog**

<!--
:cl: Onezero0
- tweak: Portstrike now happens randomly 1-3 hours into the round.


AthrunZala33092:
We need to have some indication at the join menu that portstrike is enabled, or we're going to have issues with people joining and not knowing, or people accidentally breaking the rules.

SuperGDPWYL:
PEAK

what athrun said is valid though, you need to have a way to know that portstrike is on. because right now you can only tell via the fact the 4 hour mark has passed if you join afterwards. even if it's just a note on your PDA or a special alert level a la Delta on stationforks.

LassiH:
imo it should be 2-4h mark. 1h feels too early

SuperGDPWYL:
I don't mind 1 hour. it's early enough that factions don't feel too comfortable. if it was 2 hours we might just see a culture of "okay we need to prepare for 2h" which, urgh. 1 hour is short enough that you basically have to be on your toes, which is how it should feel, because the first hour does go by quickly with basic gearing up.

AthrunZala33092:
I do think 1 hour in a 6 hour round is going to encourage even more base sitting but, at the very least we just need a notification somewhere in game so we know. Ambiguousness leads to accidental rule breaks.

Honestly the last time though we had portstrike for half the round it just became exhausting having to worry about defending the base from every Spasaka trying to break in. (And you know they will.) Are bases even strong enough to defend from the Jupiter being sent out an hour in?

(I also want a round where there is just NO portstrike save for breaking out of jail, that way people don't need to base sit at all and can go out.)

Honestly...could we change this to 2-4? Last time I had to defend Halcyon for 4 hours in an 8 hour round it was pure exhaustion. And this change varies from 3-5 hours of having to do that now.

kasature90 (changes requested):
this does need some way to determine for latejoins if its pre/post portstrike prob just simple PDA bool

kasature90 (approved):
ok nvm i just did it myself but merge that with this (and edit yours to use it instead)

kasature90:
Massive support for this in vote

PR #4412, opened 2026-08-01
W Balance

kasature90:
## About the PR
Makes every gun too OP for me to care about balance anymore.
Pistols, DMRs/snipers, and revolvers get giant damage buffs.
Rifles, LMGs, and SMGs get Realistic Firerates (so imagine like double the firerate for everything).
Shotguns get damage modifiers too (not 4ga though, and 12ga still doesnt beat it)
also gives fullauto back to wspr for w as val larps

Audio is a little bit fucky with this but you can just `cvar audio.default_concurrent 32` (or more) in console

## Why / Balance
True balance through making every weapon kill in less than a second. No more whining about 1 RPS difference.

## Requirements

- [X] I have read relevant guidelines/documentation to this PR found on [our devwiki](https://monolith-station.github.io/mono-docs/).
- [X] I have added media to this PR or it does not require an ingame showcase.
- [X] I can confirm this PR contains either no AI-generated content, or AI-generated content that meets our guidelines.

## Breaking changes

**Changelog**

:cl:
- tweak: Every weapon is like really powerful now. If you're having audio problems with the firerate, do "cvar audio.default_concurrent 32" in console.
- tweak: WSPR has full auto again.

PR #4429, opened 2026-08-02
Revert "W Balance"

OneZerooo0:
Reverts Monolith-Station/Monolith#4412

Turns out it was slop


kasature90 (approved):
insane data

tonotom1:
Try nerfing outlyer armors with beyond 50% resists but still high mobility instead (assuming there are any that still exist I haven't been able to keep up)

PR #3681, opened 2026-04-03
Research Datafarms Buffs

Redrover1760:
## About the PR

Buffs research datafarm to 20 points per second from 10.

Buffs faction datafarms to 200 points per second from 35.

## Why / Balance

A: The amount of research points a datafarm takes to make means that research datafarms made them very underused compared to crypto ones.

B: Faction gameplay really shouldn't center around research. This ensures that, for the most part, unlocking all the tech will be viable around 3 hours or so, even if no one signs up for that faction.

Ultimately, its to help work on discouraging people to camp and scislop.

## Requirements

- [X] I have read relevant guidelines/documentation to this PR found on [our devwiki](https://monolith-station.github.io/mono-docs/).
- [X] I have added media to this PR or it does not require an ingame showcase.
- [X] I can confirm this PR contains either no AI-generated content, or AI-generated content that meets our guidelines.

## Breaking changes

**Changelog**

:cl:
- tweak: Faction datafarms now produce 200 science per second instead of 35.
- tweak: Normal research datafarms now produce 20 science per second instead of 10.

PR #4432, opened 2026-08-02
Revert "Sevastopol Buff (And Faction Nerf)"

NazrinNya:
its slog because needs proper poi capture system and not whatever that disk printing shit is


OneZerooo0:
not sure about this since the sevas pointgen means factions need to contend for resources to progress

projectkepler-ru:
zero if you ABSOLUTELY insists we need to use seva just make faction point gen like 35 so that they have to atleast try. But existing so that if you late join you're not cucked by everyone being useless.

OneZerooo0 (changes requested):
ok but don't give them back the 200 point generators, keep it at like 35-50 at max

SupernoobTheN1:
I think it's better to keep sevas buff and give factions like 30 points generation at maximum as the others said

AthrunZala33092:
Yeah 35 was the old number before the data farms, and still required us to do our own science to progress. I'd be fine with returning to that old number.

NazrinNya:
ok king

waphun:
I liked the 200 research/s servers because they were a nice bargaining chip for dealing with people who didn't need money but still had things to trade, more importantly though it made it so that faction gameplay was less about sitting in base, poking machines (which also was the original reasoning given by redrover in #3681). For as fun as research and industry is, they are fundamentally frontier and advanced frontier respectively. They create hurdles for even getting basic things done. It makes sense there since there's quite literally nothing else to do on there (last I played at least). 

It was also sort of the last huzzah for this faction design principle as barks has already quit at that point and redrover quit a couple weeks later: 
faction_rework<br>
This was the reason that faction gear were the only one that had any fmc/dc values as asked in #4404.

waphun:
I was late...

PR #4495, opened 2026-08-16
Hardpoints

kasature90:
on separate branch so we can do this shit now since we have to rereview ships anyways

ported from hullrot (no specific PR since it's from rebase commit)
kinda shitty codewise
doesn't account for placing hardpoints under the guns because i dont care and it shouldn't happen in gameplay anyways (except for SRD, but the guns probably destroyed too in that case, and i'm too tired to fix that), just replace the gun fully
placeholder sprite for now


kasature90:
merging this now so ship rereviews can be updated on branch w/ it instead of waiting 2 gorzillion years for nonessential sprites

PR #4525, opened 2026-08-20
Partial Smuggling Sovl Revision

tonotom1:
## About the PR

I sprited a thing to partially replace the developer art purple crates. The tooltip for smuggling will still call it a "crate" but it doesn't concern me that much when it's probably just an .ftl change down the line. It even makes creepy sounds.

I don't think it would be right to replace *all* of the crates with the same crystal over and over so for now I've replaced just one of the types. More resprites can be added in the future.

## Why / Balance

Sovl + not developer art

## Media

image

## Requirements

## How to test

You can spawn it in manually to look at it

## Breaking changes

I didn't touch up the .ftl to account for the fact that the delivery tooltip calls everything a "crate" but whatever

**Changelog**

:cl:
- tweak: Partial smuggling crate resprite


OneZerooo0 (approved):
Thank god

PR #4528, opened 2026-08-21
Mantis Blade Buffs

lordsej:
## About the PR

nerfed structural damage
adds chem injection of 0.5u each of phlogiston and napalm to cause heat, poison and fire. totally storage in blade is 2.5 of each chem for a total of 5 hits

## Why / Balance

makes melee more viable and punishing if close range is reached. only the first 5 attacks can inject the chemicals in a fight

## Requirements

## How to test

spawn mantisblade and stab someone with armor/plate to simulate combat

## Breaking changes

N/A
**Changelog**

:cl:
- tweak: mantis blades cause flammable damage on first five strikes


kasature90 (changes requested):
incen or damage buffs but not both, this is 64 dmg/attack (multihit) plus firestacks lol

Sniperzero112:
> incen or damage buffs but not both, this is 64 dmg/attack (multihit) plus firestacks lol

Would that not make melee not worth taking the risk when a gun does more damage than a melee without the chance to get in their face?

kasature90:
> Would that not make melee not worth taking the risk when a gun does more damage than a melee without the chance to get in their face?

increase limb damage mult :trollface:

lordsej:
as per kasature90 request restored base damage back to previous and kept fire and slightly nerfed structure damage
- Slash: 14
- Piercing: 7

ready for re-review

DandiIion:
Cool change, Mantis Blade definately need help keeping up with the other melee options (especially considering the cost) and the current bonus limb damage multi has no effect untill after they are crit (therefore useless).

kasature90 (approved):

PR #4530, opened 2026-08-21
Laser Glass Diffusion Attempt #2

AthrunZala33092:
## About the PR

Attempt 2 of attempting to reduce damage of lasers that go through class with vibe coding.

From Ilya:

"if you want to try again, it should:
add DiffuseLayers to hitscan comp
OR those layers with normal raycast layers
check the index of what you hit but now also only select ents for having our normal target layaers
check how many ents with diffusion layers present are before what you hit
i don't expect you to understand this but an LLM would"

If this still bad and can't be adjusted with a minor tweak I'm not trying again and letting someone else do it. The laser stuff is more important to PDV than to me honestly.

EDIT: Nvm, Ilya says it's still bad and needs to be rewritten entirely. Since I can't code, I won't be attempting this again.

## Requirements

- [ ] I have read relevant guidelines/documentation to this PR found on [our devwiki](https://monolith-station.github.io/mono-docs/).
- [ ] I have added media to this PR or it does not require an ingame showcase.
- [ ] I can confirm this PR contains either no AI-generated content, or AI-generated content that meets our guidelines.

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Shared/Physics/CollisionGroup.cs
+++ b/Content.Shared/Physics/CollisionGroup.cs
@@ -24,7 +24,7 @@ public enum CollisionGroup
     InteractImpassable = 1 << 7, // 128 Blocks interaction/InRangeUnobstructed
     // Y dis door passable when all the others impassable / collision.
     DoorPassable       = 1 << 8, // 256 Allows door to close over top, Like blast doors over conveyors for disposals rooms/cargo.
-
+    LaserGlass = 1 << 9,
     MapGrid = MapGridHelpers.CollisionGroup, // Map grids, like shuttles. This is the actual grid itself, not the walls or other entities connected to the grid.
 
     // 32 possible groups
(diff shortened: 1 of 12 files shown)


Content.Shared/Physics/CollisionGroup.cs
Ilya246: no changes needed, this does not need another collision layer

Content.Shared/Weapons/Hitscan/Events/HitscanEvents.cs
```diff
@@ -78,6 +78,9 @@ public record struct HitscanRaycastFiredEvent
     /// Set to true the hitscan is cancelled (e.g. due to reflection).
     /// Cancelled hitscans should not apply damage or trigger follow-up effects.
     /// </summary>
+
+    /// Counts glass layers to reduce hitscan damage.
+    public int DiffuseLayers;
```
Ilya246: should be moved below HitEntities and changed into a PassedEntities hashset

PR #4571, opened 2026-08-24
Revert "Remove Special Guns"

OneZerooo0:
Reverts Monolith-Station/Monolith#4531

until we actually weaken the available options

### DIFF
--- a/Resources/Prototypes/_Mono/Loadouts/ViperGroup/role_loadouts_vipergroup.yml
+++ b/Resources/Prototypes/_Mono/Loadouts/ViperGroup/role_loadouts_vipergroup.yml
@@ -27,7 +27,7 @@
   - ContractorTrinkets
   - ContractorBureaucracy
   - NFSpeciesSpecific
-# - CivGunCaseSpecial # Mono
+  - CivGunCaseSpecial # Mono
   - ContractorArmorPlates # Mono
 
 - type: roleLoadout
@@ -59,7 +59,7 @@
   - ContractorTrinkets
   - ContractorBureaucracy
   - NFSpeciesSpecific
-# - CivGunCaseSpecial # Mono
+  - CivGunCaseSpecial # Mono
   - ContractorArmorPlates # Mono
 
 - type: roleLoadout
@@ -91,5 +91,5 @@
   - ContractorTrinkets
   - ContractorBureaucracy
   - NFSpeciesSpecific
-# - CivGunCaseSpecial # Mono
+  - CivGunCaseSpecial # Mono
   - ContractorArmorPlates # Mono
--- a/Resources/Prototypes/_NF/Loadouts/role_loadouts.yml
+++ b/Resources/Prototypes/_NF/Loadouts/role_loadouts.yml
@@ -45,7 +45,7 @@
   - MercenaryBalaclava # Mono
   - ContractorFirearm # Mono
   - ContractorMag # Mono
-# - CivGunCaseSpecial # Mono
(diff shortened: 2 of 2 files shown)


SuperGDPWYL:
so we're adding back the unhealthy playtime requirements as well as civilians not having to interact with anyone to start with a T2 faction gun?

SuperGDPWYL:
I do not think special guns, at least in the current state of "GRIND OUT FIFTY REAL-LIFE HOURS PLAYING A SPECIFIC FACTION IN MONOLITH", promote any sort of positive gameplay regardless of what tier they are or what they actually offer.

It's going to lead to low faction PQ because people are playing factions when you can tell they don't want to because they want to be able to use an MLA-73 in expeds or whatever (an issue that was already highlighted well enough by the 15 hour Captain playtime requirement to unlock TSF AI causing people to play Captain with minimal actual want to), the prospect is EXTREMELY new player unfriendly and they discourage social interaction by letting you skip it entirely and get straight to having an Annie or whatever. 

Also they encourage unhealthy playing habits. If people are complaining they grinded hours for nothing I'd consider that equivalent to someone complaining that they're getting withdrawal symptoms, and the solution is not to give them their fucking cigarettes back. This harm was caused by their existence in the first place, and keeping them in is only going to worsen the harm in the long run. Cut your losses.

tonotom1:
It's better to allow the end user imperfect means of agency and course-correct the problems than to totally restrict their options with no alternative options to make up for it

OneZerooo0:
it's not even because of the feature, its because getting rid of it means people who have already grinded it out lose the reward they worked for

SuperGDPWYL:
if that means that people don't feel the need to grind for it in the future I think that's better than keeping it so more people can feel the need to grind exorbitant amounts of hours for this bullshit in the future.

this is a bad mechanic that has a net negative effect on the game and I'm concerned that the maintainers are falling for the sunk cost fallacy regarding it.

AthrunZala33092:
Yeah, as someone who gets a roundstart Annie as a merc...screw having roundstart T2-T3 guns in merc loadout. I've seen far, far too many people grind TSF for the special guns, admit they are doing it, and contribute nothing to TSF. It should be a reward for playing, not an incentive to grind for it. After they get it, they just disappear, but while they were playing they were an active detriment to the faction. It's not even like they can't go buy a Lee Enfield from a PMC vendor right now.

Loldaddy5676:
once again said weapons should be available to faction civs, not normal civs. 

that way it would "make sense" on how they gotten it, the weapons in the special weapons loadout isnt exactly HARD to get to normal way. 

something something good citizenship rewards.


edit: if you do decide to hear out my idea, do make sure the faction specific guns go to the right civs.

Zethine07:
we had a whole balance debacle over the fact that with t2 guns and suits being roundstart TSF would always end up losing to M90 CV-32 roundstart on derelict drill every time until it was finally killed and then everything was made t1

AthrunZala33092:
I'm not against it being a reward for faction civs to have their faction's special gun.

SuperGDPWYL:
the time requirement should be like 10h at absolute most

szyefan:
Yeah unless aviable options are tweaked this shouldn't have any business getting in.
Civs shouldnt just be able to spawn with T2/3 faction guns.
T1s at best. Even marines and Rhaks can get T1s at best.

NazrinNya:
no, i wont even approve idea of giving them T1 guns, or do that only for TSF/PDV civs with their faction.

Vagrants shouldn't have special guns at all.

PR #4623, opened 2026-09-03
Good Catch

kasature90:


waphun:
huh?

kasature90:
> huh?

Forgot the metadata points (warp & latejoin)

PR #4642, opened 2026-09-05
Civ Ship Hardpoint Pass

LassiH:
## About the PR

adds hardpoints to the zeros no other changes
## Why / Balance

zeros-0

## Requirements

## Breaking changes

**Changelog**


kasature90 (changes requested):
all of these should be replaced with new ships instead (inactive authors) + it's just More Mining Ships
only keeping this open because Zeros can stay (no other big service ships really)

the main point of this refactor is to avoid having 200 unmaintained identical ships xir

LassiH:
Ill go steal the copper wiring from everything but the zeros then

kasature90 (approved):

PR #4654, opened 2026-09-06
PDV Tweaks II (Oversight Correction)

SuperGDPWYL:
## About the PR
Added chameleon headsets, backpacks and PDAs to the PDV uplink.

## Why / Balance
The headsets, bags and PDAs PDV spawn with out them as PDV, and they have no easy way to obtain PDAs, bags or headsets that _don't_. Even if Helios had an AstroVend, Plus, there's like 20 different headsets, bags and PDAs that civs can choose to spawn with, so giving a straight chameleon option for both of them is the only real way to prevent metagaming.

Robbing a civilian is still an option if you're stingy about your DC, and chameleon on Monolith is... dodgy at best so it may still end up being the superior option.

## Media
image
image

## Requirements

## How to test
spawn as PDV
check uplink Utility tab
chameleon pda, backpack and headset

**Changelog**
:cl:
- add: Chameleon PDAs, backpacks and headsets have been added to the PDV uplink.

### DIFF
--- a/Resources/Locale/en-US/_Mono/store/pdv-uplink-catalog.ftl
+++ b/Resources/Locale/en-US/_Mono/store/pdv-uplink-catalog.ftl
@@ -101,6 +101,9 @@ uplink-pirate-chameleon-pda-desc = A PDA with integrated chameleon technology, a
 uplink-pirate-chameleon-headset-name = Chameleon Headset
 uplink-pirate-chameleon-headset-desc = A headset with integrated chameleon technology, allowing you to hide your true allegiance.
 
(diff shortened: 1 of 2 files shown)


OneZerooo0 (changes requested):
Maybe just combine some of these basics into one chameleon bundle

SuperGDPWYL:
Personally I think the modularity is better, because if you got your hands on a convincing headset but not a PDA, for example, it'd suck to have to buy the whole bundle for the whole price if you only need 1 item from it

OneZerooo0 (approved):

PR #4751, opened 2026-09-22
26sep22

kasature90:
fixes fissile fuel crates being inf rads lol
fixes more safety deposit boxes bypasses


LassiH:
I heard that syringe cases can be put into storage still. unsure if this fixes it

